PR #62, opened 2015-02-04
Update readme

tfausak:
I saw [Phare's readme](https://github.com/mirego/phare/blob/v0.6/README.md) and loved how it looked. I think ours could look just as good.


tfausak:
img_0663

PR #64, opened 2015-02-05
Create a helper function for creating stoplights

tfausak:
This pull request creates a `Stoplight` helper function that acts as an alias for `Stoplight::Light.new`. This is purely for convenience and doesn't add any new functionality.

``` rb
# Before
Stoplight::Light.new('example') { nil }.run
# After
Stoplight('example') { nil }.run
```

I got the idea from [circuit_b](https://github.com/alg/circuit_b/tree/1.1#usage). I'm not convinced that `Stoplight` is the right name for this function. Some other options include:

- `stoplight`: This seems like it would often conflict with a local variable. 
- `Stoplight.light`: This has the benefit of existing within the stoplight module, but it's still a little verbose.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 This project uses [Semantic Versioning][1].
 
+- Created a convenience function for creating stoplights.
+
 ## v0.5.1 (2014-11-19)
 
 - Fixed a logic bug that incorrectly determined red lights to be yellow.
--- a/README.md
+++ b/README.md
@@ -137,7 +137,7 @@ Stoplight::Light.default_notifiers += [Stoplight::Notifier::HipChat.new(...)]
 To get started, create a stoplight:
 
 ``` rb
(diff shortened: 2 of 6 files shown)


tfausak:
I'm also not convinced this is the right functionality. I wish I could create a function that would run the light *after* any fluent builders. For example:

``` rb
magical_function('much') { :result }
# => :result
magical_function('many') { :fluent }.with_timeout(3)
# => :fluent
```

But I'm pretty sure that's impossible. So maybe this `Stoplight` function could be a shortcut for running a stoplight. You would have to switch to `Stoplight::Light.new` for the fluent builders, but you'd be able to write this:

``` rb
Stoplight('such') { :wow }
# => :wow
```

I'm not sure. That's probably not enough of a win.

tfausak:
Hey @camdez, can you look this over?

tfausak:
We discussed this in person and agreed that it seems reasonable.

PR #71, opened 2015-07-03
Added slack notifier

s1mplex:
Hey! 

My team uses Slack so I decided to extend Stoplight with another notifier. Does it look good to go?

### DIFF
--- a/lib/stoplight/notifier/slack.rb
+++ b/lib/stoplight/notifier/slack.rb
@@ -1,0 +1,27 @@
+# coding: utf-8
+
+module Stoplight
+  module Notifier
+    # @see Base
+    class Slack < Base
+      # @return [Proc]
+      attr_reader :formatter
+
+      # @return [::Slack::Notifier]
+      attr_reader :slack
+
+      # @param slack [::Slack::Notifier]
+      # @param formatter [Proc, nil]
+      def initialize(slack, formatter = nil)
+        @slack = slack
+        @formatter = formatter || Default::FORMATTER
+      end
+
+      def notify(light, from_color, to_color, error)
+        message = formatter.call(light, from_color, to_color, error)
+        slack.ping(message)
+        message
+      end
+    end
+  end
+end
--- a/stoplight.gemspec
+++ b/stoplight.gemspec
@@ -40,6 +40,7 @@ Gem::Specification.new do |gem|
     'redis' => '3.2',
     'rspec' => '3.3',
     'rubocop' => '0.32',
+    'slack-notifier' => '1.2.1',
     'timecop' => '0.7',
     'yard' => '0.8'
   }.each do |name, version|
--- a/lib/stoplight.rb
+++ b/lib/stoplight.rb
@@ -19,6 +19,7 @@ require 'stoplight/notifier'
 require 'stoplight/notifier/base'
 require 'stoplight/notifier/hip_chat'
 require 'stoplight/notifier/io'
+require 'stoplight/notifier/slack'
 
 require 'stoplight/default'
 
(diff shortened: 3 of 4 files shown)


lib/stoplight/notifier/slack.rb
```diff
@@ -0,0 +1,29 @@
...
+
+module Stoplight
+  module Notifier
+    # @see Base
+    class Slack < Base
+      # @return [Proc]
+      attr_reader :formatter
+
+      # @return [::Slack::Notifier]
+      attr_reader :slack_notifier
```
tfausak: I might prefer calling this just `slack`. What do you think?

lib/stoplight/notifier/slack.rb
```diff
@@ -0,0 +1,29 @@
...
+
+      # @param slack_notifier [::Slack::Notifier]
+      # @param formatter [Proc, nil]
+      def initialize(slack_notifier, formatter = nil)
```
tfausak: Does this need to accept any options? I see that slack-notifier has a default [channel and username](https://github.com/stevenosloan/slack-notifier/blob/1.2.1/readme.md#setting-defaults).

lib/stoplight/notifier/slack.rb
```diff
@@ -0,0 +1,29 @@
...
+        @formatter = formatter || Default::FORMATTER
+      end
+
+      def notify(light, from_color, to_color, error)
+        message = formatter.call(light, from_color, to_color, error)
+
+        slack_notifier.ping(message)
+
+        message
+      end
```
tfausak: I would prefer removing the blank lines here, like in [`Stoplight::Notifier::HipChat#notify`](https://github.com/orgsync/stoplight/blob/v1.0.0/lib/stoplight/notifier/hip_chat.rb#L36-L40).

lib/stoplight/version.rb
```diff
@@ -1,5 +1,5 @@
 # coding: utf-8
 
 module Stoplight
-  VERSION = Gem::Version.new('1.0.0')
+  VERSION = Gem::Version.new('1.1.0')
```
tfausak: I will handle bumping the version number when I do a release, like 10eae8e6f09c5fa28ff9b5903e380ef7b4e83ef9.

stoplight.gemspec
```diff
@@ -41,7 +41,8 @@ Gem::Specification.new do |gem|
     'rspec' => '3.3',
     'rubocop' => '0.32',
     'timecop' => '0.7',
-    'yard' => '0.8'
+    'yard' => '0.8',
+    'slack-notifier' => '1.2.1'
```
tfausak: This is extremely pedantic, but I like to sort dependencies by name.

tfausak:
This looks great! Thank you for opening this pull request. I left a few line notes, but this is basically good to go. The only other thing that needs to change is [the notifiers section of the readme](https://github.com/orgsync/stoplight/blob/v1.0.0/README.md#notifiers); I will do that after merging this PR.

lib/stoplight/notifier/slack.rb
s1mplex: I was thinking of passing an already configured `Slack::Notifier` object and then you don't need any options.

You can override those options for any `Slack::Notifier#ping` but I'm not sure it's that useful in Stoplight's usecase.

On the other hand this isn't hard to change so I can modify this PR. What do you think?

s1mplex:
@tfausak thanks for the feedback! I updated the PR. The only thing left to do is to decide if we want to pass options to the Notifier. Waiting for you opinion :)

lib/stoplight/notifier/slack.rb
tfausak: Ah, of course. That makes the most sense. That's what the HipChat notifier does too. Let's keep it as it is.

tfausak:
It looks great as is! Thanks again for submitting this pull request. This is a nice feature to have.

I'll update the documentation and make a new release today.

s1mplex:
:+1:

PR #74, opened 2015-07-24
Update misleading timeout documentation

### LINKED ISSUE #73: Clarify using timeout to disable automatic recovery
In the readme, it says: "Set the timeout to -1 to disable automatic recovery." Based on this, I would have expected that after enough failures it would remain red indefinitely (i.e. not automatically recover), but this is not the case. It actually goes immediately back to yellow (because it's been longer since the failure than the negative timeout).

Is this a bug or am I just reading it incorrectly?

tfausak:
Fixes #73.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,8 +209,8 @@ light.run
 # RuntimeError:
 ```
 
-The default timeout is `60` seconds. Set the timeout to `-1` to disable
-automatic recovery.
+The default timeout is `60` seconds. Set the timeout to `Float::INFINITY` to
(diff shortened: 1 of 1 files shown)


tfe:
Could mention they could use `0` for immediate recovery, in case anyone wants the previous behavior.

tfausak:
:+1: That's a good idea. You just missed the boat for this pull request, though. I'll add it in master.

tfe:
:+1:

PR #89, opened 2016-01-25
Add blacklisted errors option

mrdziuban:
I've come across a situation where I would only like one or two types of errors to change the state of the stoplight. Currently, I would have to use the `.with_allowed_errors` method to whitelist all the other errors that could potentially occur. This adds a `.with_blacklisted_errors` method that essentially does the opposite. You can specify error classes that should affect the stoplight state, and all other errors will be raised normally.

It's important to note a couple things:

- If `blacklisted_errors` is empty (which is the default), the stoplight will behave the same way it currently does
- If an error is both allowed and blacklisted, it will not affect the state of the stoplight—`allowed_errors` takes precedence over `blacklisted_errors`

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,8 @@ a while. (The yellow state corresponds to the half open state for circuit
 
 ### Custom errors
 
+##### Allowed errors
+
 Some errors shouldn't cause your stoplight to move into the red state. Usually
 these are handled elsewhere in your stack and don't represent real failures. A
 good example is `ActiveRecord::RecordNotFound`.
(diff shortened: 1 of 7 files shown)


mrdziuban:
Not sure what's up with the Rubocop build failing—none of the offenses are part of my changes here. Any ideas?

tfausak:
RuboCop is probably failing because it's using a newer version that we have tested with. Sorry about that! I'll fix the Gemfile so that can't happen. 

I think I'm in favor of this change. Allowed errors generally work well when the exception hierarchy is well formed. Unfortunately that's not always the case. Having a blacklist gives you a way to work around that. 

I'm not a fan of the term "blacklist", but I don't have any better suggestions. When compared to "allowed" errors, I immediately think "denied", but that doesn't make sense.

mrdziuban:
Thanks @tfausak!

I wasn't sure on the best term either. I started out with "forbidden" but it didn't seem right, and I ended up settling on "blacklist" because I've seen it used in other gems/libraries (in conjunction with "whitelist"). I'm happy to change it though if you prefer something else!

tfausak:
I fixed the build on master. If you merge master into this branch, it should be green. 

Maybe the best way to fix the naming issue is to rename `with_allowed_errors` to `with_whitelisted_errors`. Then the difference between them and blacklisted errors will be more apparent.

mrdziuban:
Awesome, thanks!

I can make the change to rename to `with_whitelisted_errors` if you'd like, just let me know. Are you cool with that requiring a version bump to v2.0.0, since it's a breaking change? Or would you want to alias `with_allowed_errors` to the new method to keep backward compatibility?

tfausak:
I want to release these changes, along with an alias for `with_allowed_errors`, in a minor version bump. Then a major version bump can get rid of `with_allowed_errors` in favor of `with_whitelisted_errors`. 

> Any idea why the jruby build is reporting 0% coverage?

No idea. It's not a problem. The coverage should really only run on the latest MRI build anyway.

mrdziuban:
Cool, I just pushed up the change to `with_whitelisted_errors` and an `alias_method` call to keep `with_allowed_errors`. Let me know what you think.

tfausak:
Looks great!

tfausak:
Released in [v1.3.0](https://github.com/orgsync/stoplight/releases/tag/v1.3.0).

Issue #87, opened 2016-01-08
Thread Safety


AndreasWurm:
Hey,

Stoplight with MemoryStore is not threadsafe. 
It would be awsome if stoplight would be usable in threaded ruby code.
The only shared state between Stoplights seems to be the class variables in Stoplight::Light and the DataStore.

tfausak:
I would like Stoplight to be thread safe, but I don't have a lot of experience with thread safety in Ruby. If I understand you correctly, we would have to do at least two things to make Stoplight thread safe:

- Do something with [these class variables](https://github.com/orgsync/stoplight/blob/v1.2.0/lib/stoplight/light.rb#L26-L33) in `Stoplight::Light`. 

- Do something with [the in-memory data store](https://github.com/orgsync/stoplight/blob/v1.2.0/lib/stoplight/data_store/memory.rb#L6). 

Is that correct?

AndreasWurm:
First off, a bit of clarification why we should care about threadsafety. Its not only about writing concurrent code ourselfs but another important usecase is using threaded libraries or frameworks for example just using Rails in a threaded server, like [Puma](https://github.com/puma/puma) or [Phusion Passenger](https://www.phusionpassenger.com/library/config/apache/reference/#passengerconcurrencymodel) with `PassengerConcurrencyModel thread`.

```ruby
class SomeController < ApplicationController
  def serve
    light = Stoplight('example') do
      # some interaction with active record
      # some interaction with a web service
    end
    light.run
  end
end
```

 * Regarding the class vars we should use a thread safe array for [default_notifiers](https://github.com/orgsync/stoplight/blob/v1.2.0/lib/stoplight/light.rb#L32) for the case that during initialization multiple threads append notifiers.

* regarding the memory store the easiest fix would be to synchronize all methods using a Mutex.
Also the failures array for a light is exposed and mutated in [record_failure](https://github.com/orgsync/stoplight/blob/master/lib/stoplight/data_store/memory.rb#L23), but this can also be easily mitigated. There is probably room for performance improvements over this simple approach.

I'll give it a shot during the next week and come back to you with a suggested implementation.

tfausak:
Ok great! I will be happy to review a pull request. Let me know if I can help.

tfausak:
Any movement on this?

AndreasWurm:
sorry mate havnt found the time yet, this weekend is also pretty stuffed, give me till next friday

tfausak:
No rush.

[...]

PR #92, opened 2016-02-12
Thread safety

AndreasWurm:
Pull request for issue #87


tfausak:
This looks good! I'm not sure why the build failed. Doesn't seem like a problem with this branch. 

I wonder if the thread-safe data store should be provided as an option, much like the Redis data store. That way I can avoid adding `concurrent-ruby` as a dependency, and only users that need thread safety have to pay for it. What do you think?

AndreasWurm:
well, i think thats your decision, only point i would like to raise is to remember the fallback logic, when redis fails, the Light will fallback to use Default::DATA_STORE.

tfausak:
That's a good point. Can you run [the benchmarks](https://github.com/orgsync/stoplight/blob/v1.3.0/bench/memory_bench.rb) against this branch?

tfausak:
The benchmarks don't test the failure code path, but they run in about the same amount of time as on master.

```
# master
     creating lambda:  1647203.1 i/s
      calling lambda:  1381496.4 i/s - same-ish: difference falls within error
  creating stoplight:   544997.1 i/s - 3.02x slower
   calling stoplight:   170786.9 i/s - 9.64x slower
```

```
# this branch
     creating lambda:  1607600.8 i/s
      calling lambda:  1464719.6 i/s - same-ish: difference falls within error
  creating stoplight:   555682.0 i/s - 2.89x slower
   calling stoplight:   165307.7 i/s - 9.72x slower
```

PR #94, opened 2016-03-22
Rescues all blacklisted errors

zeisler:
When blacklisting an error that inherited from Exception Class Stoplight would fail to rescue this error because it only rescues StandardError. The proposed solution is to rescue from any blacklisted_errors, in addition to StandardError.


tfausak:
:+1: Great find!

PR #95, opened 2016-03-23
I am suggesting a change to the whitelisting and blacklisting.

zeisler:
Issues with former code:
* Any whited listed errors that didn't inherit from `StandardError` were not being rescued.

Suggested Solution:
Change the whitelist to be a whitelist of `StandardError` errors.
Change the blacklist to be a blacklist of not `StandardError` errors or `Exception` errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,9 +126,6 @@ Some errors shouldn't cause your stoplight to move into the red state. Usually
 these are handled elsewhere in your stack and don't represent real failures. A
 good example is `ActiveRecord::RecordNotFound`.
 
-By default all errors that inherit from StandardError will be caught by stoplight. 
-To ignore specific errors add them to the whitelist.
-
 ``` rb
 light = Stoplight('example-not-found') { User.find(123) }
   .with_whitelisted_errors([ActiveRecord::RecordNotFound])
@@ -143,26 +140,27 @@ light.color
 # => "green"
 ```
 
+The following errors are always whitelisted: `NoMemoryError`, `SignalException`,
+`Interrupt`, and `SystemExit`.
+
 Whitelisted errors take precedence over [blacklisted errors](#blacklisted-errors).
 
 #### Blacklisted errors
 
-By default errors that inherit from `Exception`, other than `StandardError`, are not caught by Stoplight. This is because
-this would catch errors like `Interrupt`, when Ctrl-C is hit, and `NoMemoryError`. Never the less
-there may be times when you need Stoplight to catch a child of `Exception`, to do so add it to
-the blacklist.
+You may want only certain errors to cause your stoplight to move into the red
+state.
 
 ``` rb
 light = Stoplight('example-blacklist-zero') { 1 / 0 }
-  .with_blacklisted_errors([SystemExit])
+  .with_blacklisted_errors([ZeroDivisionError])
 # => #<Stoplight::Light:...>
 light.run
-# SystemExit: divided by 0
+# ZeroDivisionError: divided by 0
 light.run
-# SystemExit: divided by 0
+# ZeroDivisionError: divided by 0
 light.run
 # Switching example-blacklist-zero from green to red because ZeroDivisionError divided by 0
-# SystemExit: divided by 0
+# ZeroDivisionError: divided by 0
 light.color
 # => "red"
 ```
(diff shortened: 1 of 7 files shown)


tfausak:
This is a natural followup to #94. I'll find some time to review it soon. 

I'm not concerned about the coverage dropping by 0.2%.

README.md
```diff
@@ -125,6 +125,9 @@ Some errors shouldn't cause your stoplight to move into the red state. Usually
 these are handled elsewhere in your stack and don't represent real failures. A
 good example is `ActiveRecord::RecordNotFound`.
 
+By default all errors that inherit from StandardError will be caught by stoplight.
```
tfausak: Nits: "StandardError" should be in backticks, and "by stoplight" can be removed.

README.md
```diff
@@ -139,27 +142,26 @@ light.color
...
-The following errors are always whitelisted: `NoMemoryError`, `ScriptError`,
-`SecurityError`, `SignalException`, `SystemExit`, and `SystemStackError`.
-
 Whitelisted errors take precedence over [blacklisted errors](#blacklisted-errors).
 
 #### Blacklisted errors
 
-You may want only certain errors to cause your stoplight to move into the red
-state.
+By default errors that inherit from `Exception`, other than `StandardError`, are not caught by Stoplight. This is because
```
tfausak: :+1: I like this wording. Can you wrap this section to 80 columns?

README.md
```diff
@@ -139,27 +142,26 @@ light.color
...
+the blacklist.
 
 ``` rb
 light = Stoplight('example-blacklist-zero') { 1 / 0 }
-  .with_blacklisted_errors([ZeroDivisionError])
+  .with_blacklisted_errors([SystemExit])
 # => #<Stoplight::Light:...>
 light.run
-# ZeroDivisionError: divided by 0
+# SystemExit: divided by 0
```
tfausak: Even though `SystemExit` is blacklisted, this stoplight is still throwing a `ZeroDivisionError`.

lib/stoplight/default.rb
```diff
@@ -2,16 +2,9 @@
...
-      ScriptError,
-      SecurityError,
-      SignalException,
-      SystemExit,
-      SystemStackError
-    ].freeze
+    WHITELISTED_ERRORS = [].freeze
 
-    BLACKLISTED_ERRORS = [].freeze
+    BLACKLISTED_ERRORS = [StandardError].freeze
```
tfausak: Although there's only one error in this list, I would prefer to have it on its own like like `WHITELISTED_ERRORS` was.

spec/stoplight/default_spec.rb
```diff
@@ -28,10 +28,6 @@
       expect(Stoplight::Default::BLACKLISTED_ERRORS).to be_an(Array)
     end
 
-    it 'is empty' do
-      expect(Stoplight::Default::BLACKLISTED_ERRORS).to be_empty
-    end
```
tfausak: This spec should check `WHITELISTED_ERRORS` now. And the "contains exception classes" spec from the whitelisted errors should be run against the blacklisted errors instead.

lib/stoplight/light/runnable.rb
```diff
@@ -63,7 +63,6 @@ def not_blacklisted_error?(error)
 
       def handle_error(error, on_failure)
         raise error if whitelisted_errors.any? { |klass| error.is_a?(klass) }
-        raise error if not_blacklisted_error?(error)
```
tfausak: Why did this line go away?

zeisler: This is because any not blacklisted errors will not ever be rescued.

README.md
zeisler: Yes, this change does not make sense, let me think of a better example use case.

zeisler:
I'm not sure yet if my changes are most optimal, let me explain. It is an exceptional case to rescue `Exception` class errors because of there implications ie. `NoMemoryError`, so they should be explicitly blacklisted. In my current changes that is well established. The next area is `StandardError`, I've went back forth it can be confusing to think about a blacklist and a whitelist being used at the same time. 

I've come to the conclusion that in addition to my changes there needs to be user supplied Proc where they can decide whether Stoplight should care about an error.  This would provide a customizable interface for user and reduce code for additional use cases.

Thank you @tfausak for reviewing my PR.

tfausak:
A `Proc` would be a lot simpler. That might be the best way forward here.

zeisler:
Use RSpec's code as a battle tested choice for choosing which error to not rescue.

User can override whitelists and blacklist strategy and provide custom strategy using `#with_error_handler`

@tfausak That's for your time in reviewing this.

tfausak:
I think this will take an overhaul of the exception handling in Stoplight. The end goal here is to completely remove the whitelist and blacklist. Both will be replaced by this new `Proc`. I'll try to get a proof of concept together today.

tfausak:
Can you have a look at https://github.com/orgsync/stoplight/pull/97?

PR #97, opened 2016-04-05
Replace white/blacklists with error handler

### LINKED ISSUE #91: Remove with_allowed_errors
#89 introduced `with_whitelisted_errors` and effectively deprecated `with_allowed_errors`. The latter should be removed as part of the v2.0.0 release.

tfausak:
This will fix #91. It still needs some tests.

### DIFF
--- a/lib/stoplight/default.rb
+++ b/lib/stoplight/default.rb
@@ -2,18 +2,9 @@
 
 module Stoplight
   module Default
-    WHITELISTED_ERRORS = [
-      NoMemoryError,
-      ScriptError,
-      SecurityError,
-      SignalException,
-      SystemExit,
-      SystemStackError
-    ].freeze
-
-    BLACKLISTED_ERRORS = [].freeze
-
     DATA_STORE = DataStore::Memory.new
+
+    ERROR_HANDLER = -> (error, handler) { handler.call(error) }
 
     ERROR_NOTIFIER = -> (error) { warn error }
 
--- a/lib/stoplight/light/runnable.rb
+++ b/lib/stoplight/light/runnable.rb
@@ -52,18 +52,12 @@ module Stoplight
         failures = clear_failures
         on_success.call(failures) if on_success
         result
-      rescue *[StandardError].concat(blacklisted_errors) => error
+      rescue Exception => error # rubocop:disable Lint/RescueException
         handle_error(error, on_failure)
       end
 
-      def not_blacklisted_error?(error)
-        !blacklisted_errors.empty? &&
-          blacklisted_errors.none? { |klass| error.is_a?(klass) }
-      end
-
       def handle_error(error, on_failure)
-        raise error if whitelisted_errors.any? { |klass| error.is_a?(klass) }
-        raise error if not_blacklisted_error?(error)
+        error_handler.call(error, Error::HANDLER)
         size = record_failure(error)
         on_failure.call(size, error) if on_failure
         raise error unless fallback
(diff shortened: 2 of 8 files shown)


lib/stoplight/default.rb
```diff
@@ -2,16 +2,15 @@
...
-    BLACKLISTED_ERRORS = [].freeze
+    ERROR_HANDLER = lambda do |error|
+      classes = [
+        Interrupt,
+        NoMemoryError,
+        SignalException,
+        SystemExit
+      ]
+      raise error if classes.any? { |klass| error.is_a?(klass) }
+    end
```
zeisler: I understand your drive for simplicity, which I can see very well in this well factored project, and I think your way is better ascetically. I like the idea of removing the whitelists and blacklists completely.

Never the less this approach could lead to an API that's too raw, giving user the ability to shoot them self's in the foot. 

1. A user provided error handler will have to know about the special classes in the default handler, in some cases, in order not to rescue critical errors.
2. There could be system side effect with having to re-raise the error, especially since there are such critical errors.

I think there are some very smart people working over at RSpec and would prefer their `Module::===` method for validating what errors to rescue.
https://github.com/rspec/rspec-support/blob/master/lib/rspec/support.rb#L122

It may also make for a better `Proc` API to return `true` if the user want Stoplight to handle the issue and `false` for Stoplight to ignore it. This would take a little meta programming to create a dynamic module that takes the user supplied handler.

Sorry for changing my mind so many times.

zeisler: I will create a PR, based off your's to show in code what I mean.

lib/stoplight/light/runnable.rb
```diff
@@ -52,18 +52,12 @@ def run_code(on_success, on_failure)
         failures = clear_failures
         on_success.call(failures) if on_success
         result
-      rescue *[StandardError].concat(blacklisted_errors) => error
+      rescue Exception => error # rubocop:disable Lint/RescueException
```
zeisler: The fact that you need disable rubocop here might be a warning sign to change the API.

lib/stoplight/default.rb
tfausak: Thanks for the feedback! 

1. I agree that this is not so nice. You can fall back to the default handler as the last line of your custom handler: `return Stoplight::Default::ERROR_HANDLER.call(error)`. That's a little wordy, so people probably won't do it.

2. I don't think there's any problem with catching and re-raising errors. 

3. The difference between my method and RSpec's method is cosmetic. 

4. I considered this approach, but ultimately decided against it because it could be confusing. Does `true` mean "re-raise the error", or is that `false`? By explicitly calling `raise` in your handler, there is no confusion. Also, the only thing Stoplight would ever do to handle an error is re-raise it, so why have that extra layer of indirection?

I think the first point is the biggest problem with this approach. There may be an easy way to fix it, like returning some sentinel value to use the default handler.

lib/stoplight/light/runnable.rb
tfausak: In general I agree, but RuboCop ultimately offers suggestions. In this case, I explicitly want to catch all exceptions and then manually select some to re-raise.

lib/stoplight/default.rb
zeisler: I do agree that the `true`/`false` meaning is not apparent. What about a passed in object that responds with `#handle_error(error)` to be called in the case where the user want it to affect stoplight.

tfausak:
Once this has some tests it will be good to go.

PR #98, opened 2016-04-05
Replace white/blacklists with error hanlder and never re-raise errors

zeisler:
@tfausak Checkout this PR when you have a chance. Thanks!


zeisler:
Github is not showing the most up to date version of my branch.

zeisler:
It seem to be up-to-date now.

zeisler:
Do you have any feedback on this? @tfausak

tfausak:
I'm sorry it took me so long to look at this. I like some of what you've done here. In particular I think sending the error to a method in Stoplight (via `handler.handle` right now) is a good way to keep the sane default behavior around. I don't like creating all these modules and instances just to check a value against a function. 

This feels like a good improvement on my work in #97. I want to think a little more about how to handle the fallback case without so much `.new`.

tfausak:
I am going with #97 instead of this or #99. Thanks again for your help!

PR #99, opened 2016-04-21
Insure only rescuing safe errors

zeisler:
Remove dynamic `Module.===` but insure only rescuing safe errors
Needs test refactor and fixing.


tfausak:
I still feel like this is too complicated. See https://github.com/orgsync/stoplight/pull/97/commits/2a1261247dc16631e597da0760341867ab96ac4d for my approach.

zeisler:
@tfausak I think your commit make sense, I'll feel good how it prevent users from easily shooting them selves in the foot. I also see the performance improvements with using a Proc that gets used over and over again instead instantiating a new object every time for the error handler.

tfausak:
I appreciate the work that you put into this, and I'm glad that you helped me figure out the best way to solve this problem. That being said, I'm going to move forward with my changes in #97.

PR #102, opened 2016-04-28
Add support ActiveSupport::Testing::TimeHelpers

billthompson:
ActiveSupport::Testing::TimeHelpers stubs `Time.now` automatically, so Timecop
won't be a required dependency when testing Stoplight inside of a gem
consumer's project.

------

Thanks for making this gem! If you have any specific contributor guidelines that I haven't followed, just let me know.

### DIFF
--- a/lib/stoplight/failure.rb
+++ b/lib/stoplight/failure.rb
@@ -15,7 +15,7 @@ module Stoplight
     # @param error [Exception]
     # @return (see #initialize)
     def self.from_error(error)
-      new(error.class.name, error.message, Time.new)
+      new(error.class.name, error.message, Time.now)
     end
 
     # @param json [String]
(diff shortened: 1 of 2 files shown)


tfausak:
Thanks for opening this pull request! I'm all about keeping dependencies in check. However Stoplight doesn't depend on Active Support — not even for development. Timecop itself has no dependencies, so I'd prefer to keep using it instead of Active Support.

billthompson:
Thanks for the quick response!

I'm sorry I wasn't more clear in my description. This PR doesn't change Stoplight's dependencies or existing development flow. Notice that the specs weren't updated and all tests continue to pass.

That's because `Time.now` is the same as `Time.new` when called with no arguments. By accepting this change, consumers who use [ActiveSupport::Test::TimeHelpers](http://api.rubyonrails.org/classes/ActiveSupport/Testing/TimeHelpers.html) instead of Timecop can successfully time travel when testing.

I hope this is more clear, but please let me know if you have any questions.

tfausak:
Oh, I see. This is for other test suites that happen to use Stoplight in them. Makes total sense. 

Thanks for the patch, and I'll be sure to use `Time.now` in the future!

PR #103, opened 2016-06-01
Only clear failures if the color stayed the same

### LINKED ISSUE #93: Long running code changes red stoplight back to green too quickly
When a long-running block of code is called and the stoplight changes from green to red while that code is still running, it is possible for the failures to be cleared and the stoplight to be changed back to green as soon as the long-running code completes and `run_green` continues executing. Consider the following scenario, which describes a stoplight turning from green to red due to a database concurrency limit:

| Event | Start time | Stoplight color | Stoplight method | Resulting action |
| --- | --- | --- | --- | --- |
| Query 1 starts | 11:00:00 | Green | `run_green` | Query begins |
| Query 2 starts | 11:00:05 | Green | `run_green` | Query is rejected because of the concurrency limit; stoplight switches from green to red |
| Query 1 completes | 11:00:10 | Red | `run_green` (continued from when query 1 started) | Failures are cleared; stoplight switches from red to green |

In this case, the stoplight was only red for 5 seconds, rather than the default timeout of 60 seconds. Additionally, no notifications indicate that the stoplight returned from red to green, because `run_green` does not supply an `on_success` callback.

A possible solution here is to store the color of the stoplight before running the code and only call `clear_failures` after the code finishes executing if the stoplight color has not changed.

tfausak:
This is a work in progress. It fixes #93. 

I'm not sold on this approach yet. It makes another request to the data store. I have tried hard to keep the number of requests down. However taking the performance hit may be worth it for correctness. 

What do you think of this, @mrdziuban?

### DIFF
--- a/lib/stoplight/light/runnable.rb
+++ b/lib/stoplight/light/runnable.rb
@@ -32,14 +32,14 @@ module Stoplight
         on_failure = lambda do |size, error|
           notify(Color::GREEN, Color::RED, error) if size == threshold
         end
-        run_code(nil, on_failure)
+        run_code(Color::GREEN, nil, on_failure)
       end
 
       def run_yellow
         on_success = lambda do |failures|
           notify(Color::RED, Color::GREEN) unless failures.empty?
         end
-        run_code(on_success, nil)
+        run_code(Color::YELLOW, on_success, nil)
       end
 
       def run_red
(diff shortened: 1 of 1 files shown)


tfausak:
Closing due to lack of interest.

PR #106, opened 2016-08-03
Add Sentry/Raven notifier

knapo:
Similar to Honeybadger, would be nice to have Sentry/Raven notifier.


tfausak:
Looks great. Thank you!

PR #108, opened 2016-09-22
Add timeout option

GabrielMalakias:
I like this gem and trying to use in some projects! Good work community :)

Some days ago, i've found a interesting property in Hystrix:
"execution.isolation.thread.timeoutInMilliseconds

This property sets the time in milliseconds after which the caller will observe a timeout and walk away from the command execution."

I now, this gem isn't Hystrix, but adding timeout limit can you avoid latency (in some cases) and cascading failures. 

I suggest this code, but it's only an opinion of a paltry dev :)

Thanks for all.

### DIFF
--- a/lib/stoplight/default.rb
+++ b/lib/stoplight/default.rb
@@ -23,5 +23,7 @@ module Stoplight
     ].freeze
 
     THRESHOLD = 3
+
+    TIMEOUT = 1000
   end
 end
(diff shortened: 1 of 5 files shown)


tfausak:
I'm glad you like this gem, and thanks for opening this pull request! 

Stoplight doesn't have timeouts because, as you've shown, they're very easy to add. I would rather keep Stoplight focused on doing one thing well. If you want something a little closer to Hystrix, I suggest checking out [CircuitBox](https://github.com/yammer/circuitbox).

GabrielMalakias:
Thanks Taylor. :)

PR #109, opened 2017-01-13
Switch hard coded state key

AHaymond:


tfausak:
Cool. Looks good. Thanks!

PR #111, opened 2017-02-07
Adds time format constant and some changees at from_json method

GabrielMalakias:


tfausak:
Thanks for opening this PR! You can ignore the build failure; Rubinius is having trouble. 

Can you explain your motivation for making these changes?

GabrielMalakias:
Ok, well I saw this unfrozen string then I checked if Ruby already has a predefined constant to use it but hasn't. I defined the time_format constant to avoid string creation when executes from_json. After that I created an object to avoid access ['error'] in hash two times. 

I decided do this pull request because I'm using it then I see this to do improvements. It hasn't a big thing but is an improvement, I think.

spec/stoplight/default_spec.rb
```diff
@@ -77,4 +77,14 @@
...
     end
   end
+
+  describe '::TIME_FORMAT' do
+    it 'is a string' do
+      expect(Stoplight::Default::TIME_FORMAT).to be_a(String)
+    end
+
+    it 'is frozen' do
+      expect(Stoplight::Default::NOTIFIERS).to be_frozen
```
tfausak: This is checking the wrong value.

lib/stoplight/default.rb
```diff
@@ -23,5 +23,7 @@ module Default
     ].freeze
 
     THRESHOLD = 3
+
+    TIME_FORMAT = '%Y-%m-%dT%H:%M:%S.%N%:z'.freeze
```
tfausak: The `Default` module is meant for things that can be changed by the user. Would you mind moving this constant into the `Failure` module?

spec/stoplight/default_spec.rb
GabrielMalakias: Sorry.. :)

lib/stoplight/default.rb
GabrielMalakias: Ok, I'm going to move it. Thanks!

PR #115, opened 2018-03-28
Support light names containing colons

f-mer:
Light names containing colons lead to names returned by `data_store.names` to be incomplete
eg. `Stoplight('http://api.example.com/some/action')` and `names` will return `['//api.example.com/some/action']`. 

This leads to a misleading representation of the current state on [stoplight-admin](https://github.com/orgsync/stoplight-admin/blob/master/lib/sinatra/stoplight_admin.rb#L24).


f-mer:
Thans for your quick response! :smiley:
It fixes the described bug in spotlight-admin. :tada:

PR #119, opened 2018-09-08
Replace Fixnum with Integer in default_spec.rb to eliminate deprecation warning

mkrogemann:
Replacing Fixnum with Integer removes a deprecation warning that you would receive with MRI versions 2.4 and 2.5 respectively but also works fine for 2.1-2.3 versions so I consider this a safe improvement.
I also added 2.4 and 2.5 to the build matrix for travis ci. Also added some indentation to the .travis.yml file where my editor felt that it was in order to do so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,11 +7,13 @@ cache: bundler
 language: ruby
 matrix:
   include:
-  - env: RAKE_TASK=rubocop
-    rvm: '2.3.0'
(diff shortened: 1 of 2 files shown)


tfausak:
Thanks for opening this pull request! Unfortunately this repository is not maintained. If you're interested in becoming the maintainer, let me know. Otherwise I recommend you use your fork.

PR #132, opened 2020-01-12
[WIP] Add threshold with sliding window

### LINKED ISSUE #131: Failures counted by unit of time
## Context

## Problem

## Proposal

bolshakov:
Fixes #131

And there are several unhandled issues left which must be resolved before merging:

* [ ] Carefully cover new code with tests. 
* [ ] Do not break backward compatibility (The new algorithm most likely should use a separate namespace for keys)
* [ ] Get rid of warnings.
* [ ] Update docs.

### DIFF
--- a/lib/stoplight/data_store/redis.rb
+++ b/lib/stoplight/data_store/redis.rb
@@ -40,10 +40,23 @@ module Stoplight
         normalize_failures(query_failures(light), light.error_notifier)
       end
 
+      WINDOW_SIZE = Float::INFINITY
+
+      #  TODO: time in utc
+      #  TODO: investigate warnings
+      #  TODO: tests with different window sizes
       def record_failure(light, failure)
-        size, = @redis.multi do
-          @redis.lpush(failures_key(light), failure.to_json)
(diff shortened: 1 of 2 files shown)


lib/stoplight/data_store/redis.rb
```diff
@@ -40,10 +40,23 @@ def get_failures(light)
...
-          @redis.ltrim(failures_key(light), 0, light.threshold - 1)
+        # @redis.lpush(failures_key(light), failure.to_json)
+        # @redis.ltrim(failures_key(light), 0, light.threshold - 1)
+        failures_key = failures_key(light)
+
+        _, size, = @redis.multi do
+          @redis.zadd(failures_key, failure.time.to_i, failure.to_json)
+          @redis.zcard(failures_key)
+          # keep at most WINDOW_SIZE old errors
+          @redis.zremrangebyscore(failures_key, 0, Time.new.to_i - WINDOW_SIZE)
```
thebigw4lrus: This is awesome.. Can I help you with this, this is super useful to me, and I need it asap!. thanks!

bolshakov:
Closed in favor of https://github.com/orgsync/stoplight/pull/133

PR #133, opened 2020-02-23
WIP: Feature/add window size

### LINKED ISSUE #131: Failures counted by unit of time

thebigw4lrus:
Fixes #131

And there are several unhandled issues left which must be resolved before merging:

* [x] Carefully cover new code with tests. 
* [ ] Do not break backward compatibility (The new algorithm most likely should use a separate namespace for keys)
* [x] Update docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -203,6 +203,32 @@ light.run
 
 The default threshold is `3`.
 
+### Custom window size
+
+By default, all recorded failures regardless the time these happens will count to reach
+the threshold(hence to turn the light to red). In case if needed, a window size can be set,
+meaning you can control how many errors per period of time will count to reach the red
+state.
+
+``` rb
+window_size_in_seconds = 2
+
+light = Stoplight('example-threshold') { 1 / 0 }
+  .with_window_size(window_size_in_seconds)
+  .with_threshold(1)
+# => #<Stoplight::Light:...>
+
+light.run
+# #<ZeroDivisionError: divided by 0>
+# => "default"
+sleep(3)
+light.run
+# #<ZeroDivisionError: divided by 0>
+# => "default"
+```
+
+The default window size is infinity, so all failures counts.
+
 ### Custom cool off time
 
 Stoplights will automatically attempt to recover after a certain amount of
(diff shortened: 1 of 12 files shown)


thebigw4lrus:
@bolshakov Can you take a look to this?.. one thing I havent managed to resolved are the warnings.  It seems it comes from the Hash ruby implementation used by fakeredis.

thebigw4lrus:
@bolshakov I really need this 🙏 ... can you take a look? thanks!.

bolshakov:
Hey, @thebigw4lrus thank you for your patience and effort. I apologize for my absence. 

You've done a great job! 👍

Do not worry about warnings👌. The main concern now is backward compatibility. Imagine you're switching from the current version to a new one that uses Sorted Sets instead of lists.  Thus your `failures_key(light) ` is not empty. You'll get an error when one any attempt to perform Sorted Set commands on it:

```redis
127.0.0.1:6379> lpush foo bar
(integer) 2
127.0.0.1:6379> zadd foo 42 baz
(error) WRONGTYPE Operation against a key holding the wrong kind of value
```

It seems the only option is to alter between old and new behavior depending on a `failure_key` type.  What do you think?

thebigw4lrus:
Hi @bolshakov !. Yeah, we could go that way!.. just wondering if one have a declared light, and upgrade the gem, it will only go for the former behaviour right?(with no window size), hence, that person would never be able to benefit from this new feature right?. 

How about to introduce a breaking version? (so all users would lost their recorded failures if they upgrade to the next version, but then we would have all users in the same library state).  Maybe this could make sense in order to avoid backward compatible adjustment code.  At the end, the gem upgrade can be done when the circuit remains closed.

What do you think?

bolshakov:
@thebigw4lrus I experimented a bit and probably the simpler and more convenient to migrate keys from Lists to ZSets on first use. I'm working in that direction now

thebigw4lrus:
Amazing!!! many thanks @bolshakov !. Let me know if I can help with such move.

bolshakov:
@thebigw4lrus, thank you for your work ❤️. It took some time, but I merged your changes into the master branch. I started a different pr for that https://github.com/bolshakov/stoplight/pull/172

I'm going to include these changes in version 4.0

PR #136, opened 2020-11-12
Update links to the correct repo org

tasomaniac:


tfausak:
I feel like this repo should really live under @bolshakov's namespace. It's not associated with OrgSync or Campus Labs anymore.

bolshakov:
Yeah, I agree!

bolshakov:
@tasomaniac thank you for pointing out the problem! As @tfausak suggested, I moved the repo. I'll update links and other things by myself later this weekend.

PR #146, opened 2021-01-27
Test against ruby 3.0.x

bolshakov:


jxa:
Took a swing at the ruby 3 specs https://github.com/bolshakov/stoplight/pull/148

PR #148, opened 2022-01-10
Update gems to pass ruby 3 build

jxa:
👋 Howdy! Specs pass locally with ruby 3 by updating a couple of gems. I executed the following and committed the resulting gemfile.lock

```shell
bundle update mimemagic
bundle update simplecov
```


bolshakov:
Thank you for your contribution!

PR #150, opened 2022-01-25
Fix deprecated uses of `Redis#pipelined`

casperisfine:
Context: https://github.com/redis/redis-rb/pull/1059

The following is deprecated
```ruby
redis.pipelined do
  redis.get(key)
end
```

And should be rewritten as:
```ruby
redis.pipelined do |pipeline|
  pipeline.get(key)
end
```

Functionally it makes no difference.
This API is available since Redis 3.0.

### DIFF
--- a/lib/stoplight/data_store/redis.rb
+++ b/lib/stoplight/data_store/redis.rb
@@ -25,9 +25,9 @@ module Stoplight
       end
 
       def get_all(light)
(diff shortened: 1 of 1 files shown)


mrzasa (approved):

bolshakov:
Thank you for the contribution ❤️

Issue #162, opened 2022-12-03
Lock light state using the Light class interface


Lokideos:
To manually lock the state of light, we use the `data_store` interface. For example:
```ruby
light = Stoplight('example-locked') { true }
# => #<Stoplight::Light:..>
light.run
# => true
light.data_store.set_state(light, Stoplight::State::LOCKED_RED)
# => "locked_red"
light.run
# Stoplight::Error::RedLight: example-locked
```

However, it allows us to use the interface of one light to lock the state of any other light. To avoid this situation, I'd like to add the `Light#set_state` method, which will be called like this:
```ruby
light = Stoplight('example-locked') { true }
light.set_state(Stoplight::State::LOCKED_RED)
```

This implementation will clearly show for what light we lock the state.

bolshakov:
Maybe instead of emphasizing setting state, we can emphasize locking the circuit breaker:

```ruby 
light = Stoplight('example-locked') { true }
light.lock(Stoplight::Color::RED) 
```

Users don't have to know about the `Stoplight::State::LOCKED_RED` which is internal details aimed to distinguish red statute from locked red. What do you think?

Lokideos:
@bolshakov yes, that's a great idea! I also think that to unlock it, we can just use the `Light#unlock` method without any arguments. WDYT?

bolshakov:
This makes sense to me

PR #170, opened 2022-12-10
feat | Light lock interface improvement

Lokideos:
This ticket aims to improve the interface of the locking light state as described in https://github.com/bolshakov/stoplight/issues/162.

Previously, if we needed to lock the light state, we used the data store interface instead of the light interface.
This approach exposed the implementation of light locking and potentially allowed the locking state of one light via the interface of another light.
Now we call `#lock` directly on the light and also hide the implementation of light locking.
```ruby
# was
light.data_store.set_state(light, Stoplight::State::LOCKED_RED)
# now
light.lock(Stoplight::Color::RED)
```
We also introduce the `#unlock` method, which hides the implementation of unlocking the light and simplifies light unlocking:
```ruby
# was
light.data_store.set_state(light, Stoplight::State::UNLOCKED)
# now
light.unlock
```


lib/stoplight/light/runnable.rb
```diff
@@ -26,6 +27,21 @@ def run
         end
       end
 
+      def lock(color)
+        raise Error::IncorrectColor unless Color::LOCKABLE_COLORS.include?(color)
+
+        state = case color
+                when Color::RED then State::LOCKED_RED
+                when Color::GREEN then State::LOCKED_GREEN
+                end
```
bolshakov, line 36: What do you think? 

```suggestion       
        state = case color
                when Color::RED then State::LOCKED_RED
                when Color::GREEN then State::LOCKED_GREEN
                else raise Error::IncorrectColor
                end
```

1. you don't need the `LOCKABLE_COLORS` constant anymore
2. You don't have to synchronize the case statement with the `LOCKABLE_COLORS` constant

bolshakov (dismissed):
Great job, @Lokideos 👏 

I left a suggestion that I believe could simplify the code

lib/stoplight/light/runnable.rb
Lokideos: I think it's cool. I'll implement it right away :)

bolshakov (approved):

lib/stoplight/light/runnable.rb
```diff
@@ -26,6 +27,20 @@ def run
         end
       end
 
+      def lock(color)
```
bolshakov, line 30: Sorry, one final thought. Would you mind adding yard docs to the method? Since it's a public integrate, let's document what it returns, I suggest returning either self of color. I lean toward a color, but don't have a strong opinion about that

lib/stoplight/light/runnable.rb
```diff
@@ -26,6 +27,20 @@ def run
...
+                when Color::GREEN then State::LOCKED_GREEN
+                else raise Error::IncorrectColor
+                end
+
+        safely { data_store.set_state(self, state) }
+      end
+
+      def unlock
+        safely { data_store.set_state(self, Stoplight::State::UNLOCKED) }
+      end
```
bolshakov, line 42: On the other hand, we cannot return a color in this method 🤔

Lokideos: I'll think about what to return then :)

Lokideos: @bolshakov, because we can't return color in this method think that we should return Light here.

lib/stoplight/light/runnable.rb
Lokideos: I decided to return Light.

bolshakov (approved):

PR #171, opened 2022-12-10
Replace data store uses with strategy

bolshakov:
This second refactoring step aimed to introduce the running window behavior for counting errors. 

The `Stoplight::Strategy::Base:` is the interface that each strategy should implement. The interface is the same as  `Stoplight::DataStore::Base`.

The `Stoplight::Strategy::Vintage` implements the old-fashioned strategy that Stoplight implements today. 

All the `Stoplight::Light#data_store` usages were replaced with `Stoplight::Light#strategy` usages. 

This PR also introduces a breaking change: the public `Stoplight#data_store` accessor is not available anymore (which may affect end users, e.g., stoplight-admin) 

The PR targets the [feature/running-window](https://github.com/bolshakov/stoplight/tree/feature/running-window) branch and I suggest accepting this breaking change for now. Before the release, we may reconsider this decision

### DIFF
--- a/lib/stoplight.rb
+++ b/lib/stoplight.rb
@@ -15,6 +15,9 @@ require 'stoplight/data_store/base'
 require 'stoplight/data_store/memory'
 require 'stoplight/data_store/redis'
 
+require 'stoplight/strategy/base'
+require 'stoplight/strategy/vintage'
+
 require 'stoplight/notifier'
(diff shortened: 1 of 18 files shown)


Lokideos (approved):
Great job!
I'm not sure about naming, but we can come back to the naming issue later.

PR #172, opened 2022-12-10
Implement window size

### LINKED ISSUE #131: Failures counted by unit of time

bolshakov:
By default, all recorded failures, regardless of the time these happen, will count to reach
the threshold (hence turning the light to red). If needed, a window size can be set,
meaning you can control how many errors per period of time will count to reach the red
state.

 ``` rb
 window_size_in_seconds = 2

 light = Stoplight('example-threshold') { 1 / 0 }
   .with_window_size(window_size_in_seconds)
   .with_threshold(1)
 # => #<Stoplight::Light:...>

 light.run
 # #<ZeroDivisionError: divided by 0>
 # => "default"
 sleep(3)
 light.run
 # #<ZeroDivisionError: divided by 0>
 # => "default"
 ```

Closes #131 
Closes https://github.com/bolshakov/stoplight/pull/133


bolshakov:
@Lokideos 

Jimmeh

lib/stoplight/data_store/redis.rb
```diff
@@ -93,6 +96,17 @@ def with_notification_lock(light, from_color, to_color)
 
       private
 
+      # @param light [Stoplight::Light]
+      # @param time [Time]
+      def remove_outdated_failures(light, time, transaction: @redis)
+        failures_key = failures_key(light)
+
+        # Remove all errors happened before the window start
+        transaction.zremrangebyscore(failures_key, 0, time.to_i - light.window_size)
```
Lokideos, line 105: That's a cool idea!

lib/stoplight/data_store/redis.rb
```diff
@@ -108,7 +122,9 @@ def set_last_notification(light, from_color, to_color)
       end
 
       def query_failures(light, transaction: @redis)
-        transaction.lrange(failures_key(light), 0, -1)
+        window_start = Time.now.to_i - light.window_size
+
+        transaction.zrange(failures_key(light), Float::INFINITY, window_start, rev: true, by_score: true)
```
Lokideos, line 127: I'm not quite sure how it should work here. Should it me something like this?
```ruby
transaction.zrange(failures_key(light), window_start, light.window_size, rev: true, by_score: true)
```

lib/stoplight/data_store/redis.rb
```diff
@@ -93,6 +96,17 @@ def with_notification_lock(light, from_color, to_color)
...
 
+      # @param light [Stoplight::Light]
+      # @param time [Time]
+      def remove_outdated_failures(light, time, transaction: @redis)
+        failures_key = failures_key(light)
+
+        # Remove all errors happened before the window start
+        transaction.zremrangebyscore(failures_key, 0, time.to_i - light.window_size)
+        # Keep at most +light.threshold+ number of errors
+        transaction.zremrangebyrank(failures_key, 0, -light.threshold - 1)
```
Lokideos, line 107: Why do we want to keep at most `light.threshold` errors?

Lokideos (approved):
We should mention that now we only support Redis starting from `6.2.0` and probably change the version of Redis we use for tests to that version (or later) since new functionality will not work with Redis < `6.2.0`.
Otherwise, but I don't understand some parts of it.

lib/stoplight/data_store/memory.rb
```diff
@@ -21,18 +21,22 @@ def names
...
       end
 
       def get_failures(light)
-        synchronize { @failures[light.name] }
+        synchronize { query_failures(light) }
       end
 
       def record_failure(light, failure)
         synchronize do
           n = light.threshold - 1
```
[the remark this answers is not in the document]
bolshakov: I'll get rid of this variable at all

lib/stoplight/data_store/redis.rb
```diff
@@ -44,9 +44,12 @@ def get_failures(light)
       end
 
       def record_failure(light, failure)
-        size, = @redis.multi do |transaction|
-          transaction.lpush(failures_key(light), failure.to_json)
-          transaction.ltrim(failures_key(light), 0, light.threshold - 1)
+        *, size = @redis.multi do |transaction|
+          failures_key = failures_key(light)
```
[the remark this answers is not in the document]
bolshakov: I decided to keep it as is, but added a comment to the `failures_key` method, explaining what's the `failures_key`.

I also added class level documentation to explain the way we store erorrs.

lib/stoplight/data_store/redis.rb
bolshakov: I added a class-level comment describing this. Long story short: Since we use a threshold to trigger notifications, it does not matter if we have more failures than the threshold value. So every time we save a new failure, we ensure we don't store old failures that could occupy a lot of Redis space and put pressure on the `get_failures` method.

lib/stoplight/data_store/redis.rb
bolshakov: In the previous stoplight version, we used a List to store failures. The most recent failures were at the head of the list. 

Now we store errors in a sorted set, and we sort them by time. The most recent failures at the end of the set. 

To maintain the same interface (return recent errors at the beginning) I use `rev: true` options here. In combination with `by_score: true` it performs a reverse search by error time. According to the [documentation](https://redis.io/commands/zrange/#reverse-ranges), if you perform a reverse search you put the beginning of the range as a first argument: For instance:

```
ZRANGE zset 10 5 REV BYSCORE
```

> Will return all elements with scores less than 10 and greater than 5.

bolshakov:
Good point. I added tests for different Redis versions https://github.com/bolshakov/stoplight/pull/179/files
I guess we sorted out how `rev` option works 😉

PR #174, opened 2022-12-11
Drop branded notifiers

### LINKED ISSUE #163: The future of notifiers
Maintaining all the possible notifiers is hardly possible. The underlying libraries change interfaces, get outdated, or discontinued.

## Proposal

Let's drop all the notifiers from the `stoplight` gem except `io` and `logger` and let the community maintain the rest as separate gems. I can maintain the Sentry notifier.

bolshakov:
It's too much of a burden to support all possible notifiers. We decided to drop all the notifiers that require external gems and let the community support them.

This should allow having an independent release cycle for such notifiers and, therefore, better support.

Closes #163


spec/stoplight/notifier/generic_spec.rb
```diff
@@ -47,4 +47,24 @@
...
+    let(:to_color) { Stoplight::Color::RED }
+    let(:notifier) { notifier_class.new(double.as_null_object) }
+    let(:error) { nil }
+    let(:notifier_class) do
+      Class.new do
+        include Stoplight::Notifier::Generic
+      end
+    end
+
+    it 'has to implement the #put method' do
```
Lokideos, line 64: We can also test that we call the `formatter#call` method with correct arguments and return a formatted message from the `#notify` method, but it's totally optional.

Lokideos (approved):
Looks cool!
We are planning to right a separate gem for Sentry support, right?

bolshakov:
Yes, we will. I added [this ticket](https://github.com/bolshakov/stoplight/issues/178) to the roadmap https://github.com/bolshakov/stoplight/milestone/5

spec/stoplight/notifier/generic_spec.rb
bolshakov: Agreed

PR #180, opened 2022-12-20
Interface rework

### LINKED ISSUE #161: Improve Stoplight builder interface
Currently, it's not possible to build an instance of a Stoplight and pass it as a dependency because the code protected by the circuit breaker is coupled with the circuit breaker itself:

```ruby
circuit_breaker = Stoplight('api') { call_api }
  .with_threshold(42)
  .with_cool_off_time(2)
```

What if we rethink this approach and pass the code block to the `run` method instead?

```ruby
circuit_breaker = Stoplight('api')
  .with_threshold(42)
  .with_cool_off_time(2)
```

Thin enables us to pass Stoplight instances as dependencies and share them across the codebase:

```ruby
def call(circuit_breaker)
  circuit_breaker.run('api') { call_api }
end
```

This may require moving error handling and fallback out of the builder interface:

```ruby
circuit_breaker
   .with_fallback { |e| p e; 'default' }
   .with_error_handler do |error, handle|
     raise error if error.is_a?(ActiveRecord::RecordNotFound)
     handle.call(error)
  end
 .run { call_api }
```

The main disadvantage here is breaking compatibility with the previous interface, but probably it's not necessary, and we can keep both. 

⚠️ If we decide to allow both approaches, we have to make sure it won't be possible to mix them and get confused.

bolshakov:
Closes #161

**TODO**
- [x] Specs for `Configuration`
- [ ] Shell we run all specs on `Configuration` and `Light`?
- [ ] Documentation https://github.com/bolshakov/stoplight/issues/188
- [ ] Upgrade guide https://github.com/bolshakov/stoplight/issues/188
- [ ] A comment about deprecated methods https://github.com/bolshakov/stoplight/issues/188

Currently, it's impossible to build an instance of a Stoplight and pass it as a dependency. It happens because the code protected by the circuit breaker is coupled with the circuit breaker itself:

```ruby
circuit_breaker = Stoplight('api') { call_api }
  .with_threshold(42)
  .with_cool_off_time(2)
```

This PR rethinks the Stoplight's interface and enables us to pass the code block to the `run` method instead.

```ruby
circuit_breaker = Stoplight('api')
  .with_threshold(42)
  .with_cool_off_time(2)
```

Thin allows us to pass Stoplight instances as dependencies and share them across the codebase:

```ruby
def call(circuit_breaker)
  circuit_breaker.run { call_api }
end
```

The following methods have been moved to the new `Stoplight::Configuration` object:

* `#with_data_store`
* `#with_cool_off_time`
* `#with_threshold`
* `#with_window_size`
* `#with_notifiers` 

Thus, you can configure your light and pass this configuration around:

```ruby
circuit_breaker = Stoplight('api')
  .with_threshold(42)
  .with_cool_off_time(2)
  .with_data_store(Stoplight::DataStore::Memory.new)
  .with_window_size(3600)
  .with_notifiers([Stoplight::Notifier::IO.new(StringIO.new)])
```

This configuration could be reused between lights:

```ruby
circuit_breaker.with_default([]).run { get_items }
circuit_breaker.with_default(nil).run { update_items }
```

Passing a code block into the `Stoplight()` method is still possible. In that case, you'll get a `Light` instance (instead of a `Configuration` instance)

```ruby
Stoplight('api') { do_the_job }
  .with_threshold(42)
  .with_cool_off_time(2)
  .with_data_store(Stoplight::DataStore::Memory.new)
  .with_window_size(3600)
  .with_notifiers([Stoplight::Notifier::IO.new(StringIO.new)])
  .run
```

This change does not require users to change their existing code. However, the new interface behaves slightly differently:

**Old interface**: 

```ruby
circuit_breaker = Stoplight('api') { do_the_job } # => Stoplight::Light
circuit_breaker.with_threshold(42)  # => Stoplight::Light
circuit_breaker.run
```

**New interface**: 

```ruby
circuit_breaker = Stoplight('api') # => Stoplight::Configuration
circuit_breaker.with_threshold(42) # => Stoplight::Configuration
circuit_breaker.run { do_the_job }
```

The above code does not work as before because the `#with_threshold` method returns a new instance of stoplight configuration. It might be even more confusing that this works with the new interface:

```ruby
circuit_breaker = Stoplight('api').with_fallback(42) # => Stoplight::Light
circuit_breaker.with_error_notifier { |error|  ... } # => Stoplight::Light
circuit_breaker.run { do_the_job }
```

This also works:

```ruby
circuit_breaker = Stoplight('api').with_fallback(42) # => Stoplight::Light
circuit_breaker.with_threshold(42) # => Stoplight::Light
circuit_breaker.run { do_the_job }
```

It happens because `#with_data_store`, `#with_cool_off_time`, `#with_threshold`, `#with_window_size`, and `#with_notifiers` methods are defined on the configuration level. The `with_error_handler`, `with_error_notifier`, and `with_fallback` are defined on the `Light` level and actually modify the light. Those three methods are also defined on the configuration level and return a new instance of `Light` configured with the current configuration.

Thus, as soon as you call a method that returns an instance of `Stoplight:::Light`, you can modify the configuration in place.

I imagine this might not be clear to users. However, I don't see real live scenarios when this might be a problem. 

Another four methods have been moved to the `Configuration`; calling them on the `Light` class will produce a warning 
* `.default_data_store` and `.default_data_store=`
* `.default_notifiers` and `.default_notifiers=`

Opinions are welcomed.


bolshakov:
I found a few real-world examples that could benefit from the new interface:

### 1. [mastodon](https://github.com/mastodon/mastodon/blob/main/app/workers/activitypub/delivery_worker.rb#L50-L64)

They assign the circuit breaker to a variable to avoid chaining methods with a block:

```ruby
def perform_request
  light = Stoplight(@inbox_url) do
    request_pool.with(@host) do |http_client|
      build_request(http_client).perform do |response|
        raise Mastodon::UnexpectedResponseError, response unless response_successful?(response) || response_error_unsalvageable?(response)

        @performed = true
      end
    end
  end

  light.with_threshold(STOPLIGHT_FAILURE_THRESHOLD)
    .with_cool_off_time(STOPLIGHT_COOLDOWN)
    .run
end

```

With the new interface, they could do it without an intermediate variable:

```ruby
def perform_request
  Stoplight(@inbox_url)
    .with_threshold(STOPLIGHT_FAILURE_THRESHOLD)
    .with_cool_off_time(STOPLIGHT_COOLDOWN)
    .run do
      request_pool.with(@host) do |http_client|
        build_request(http_client).perform do |response|
          raise Mastodon::UnexpectedResponseError, response unless response_successful?(response) || response_error_unsalvageable?(response)
  
          @performed = true
        end
      end
    end
end
```

### 2. [mastodon-twitter-poster](https://github.com/renatolond/mastodon-twitter-poster)

In [one place](https://github.com/renatolond/mastodon-twitter-poster/blob/main/lib/mastodon_user_processor.rb#L58-L68) they wrap requests with Stoplight 

```ruby
def self.stoplight_wrap_request(domain, &)
  if domain.present?
    Stoplight("source:#{domain}", &)
      .with_threshold(3)
      .with_cool_off_time(5.minutes.seconds)
      .with_error_handler { |error, handle| error.is_a?(HTTP::Error) || error.is_a?(OpenSSL::SSL::SSLError) || error.is_a?(Oj::ParseError) ? handle.call(error) : raise(error) }
      .run
  else
    yield
  end
end
```

And in [another place](https://github.com/renatolond/mastodon-twitter-poster/blob/main/app/controllers/home_controller.rb#L6-L7) they display domain status:

```ruby
def index
  if current_user
    @uid = current_user.mastodon&.uid
    @handle = current_user.twitter_handle
    # TODO: Merge this Stoplight with the one in MastodonUserProcessor to avoid different params
    @stoplight_domain_status = Stoplight("source:#{current_user.mastodon_domain}").with_threshold(3).with_cool_off_time(5.minutes.seconds).color if current_user.mastodon
  end
end
```

They need two share stoplight configurations between these two places to display the status correctly. They can also benefit from the new interface because it allows sharing an instance of stoplight.

lib/stoplight.rb
```diff
@@ -24,11 +24,16 @@ module Stoplight # rubocop:disable Style/Documentation
 
 require 'stoplight/default'
 
+require 'stoplight/configuration'
 require 'stoplight/light/lockable'
 require 'stoplight/light/runnable'
 require 'stoplight/light'
 
 # @see Stoplight::Light#initialize
```
Lokideos, line 32: Since we now allow to return either Light or Configuration object from this method, I'd fix the documentation here.

lib/stoplight/configuration.rb
```diff
@@ -0,0 +1,135 @@
...
+      @data_store = data_store
+      @notifiers = notifiers
+      @cool_off_time = cool_off_time
+      @threshold = threshold
+      @window_size = window_size
+    end
+
+    # @param name [String]
+    # @return [Stoplight::Configuration]
+    def with_name(name)
```
Lokideos, line 72: Where do we use this private method?

lib/stoplight/configuration.rb
```diff
@@ -0,0 +1,135 @@
+# frozen_string_literal: true
+
+require 'forwardable'
+
+module Stoplight
+  # An immutable +Stoplight::Light+ configuration object that could be
```
Lokideos, line 6: I think that it would be better if we mention that we return the `Stoplight::Configuration` object here.

spec/stoplight/configuration_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+
+  describe '#run' do
+    it 'yields the block' do
+      expect do |code|
+        configuration.run(&code)
+      end.to yield_control
+    end
+  end
+
+  describe '#luck, #unlock, #color' do
```
Lokideos, line 110: I think it's lock
```suggestion
  describe '#lock, #unlock, #color' do
```

Lokideos, line 110: Also, I think that we should split this describe to three different describe blocks. One for each method.

Lokideos (commented):
Great work!
From the user's perspective it works great and allows us to comfortably use different DI approaches.
However, I took another look on how `Configuration` object is implemented and it looks like we can do better here.
Currently, if we use the new interface, we actually use `Stoplight::Configuration`, which either handles method calls himself or delegates those method calls to `Light` object. I think that it's too much responsibility for the `Configuration` object.
IMHO `Light` should know everything it needs about the `Configuration` object, but `Configuration` object should know nothing about the `Light` object. It will keep our codebase cleaner and will prevent us from issues with future `Stoplight` improvements.
Finally, we might want to write some examples of usage for our users using this new syntax. At least one example for Rails (e.g., we can create all needed configuration in a separate `Stoplight` initializer and use it later).
WDYT?

spec/stoplight/configuration_spec.rb
bolshakov: How would you suggest testing them independently from each other?

lib/stoplight/configuration.rb
bolshakov: I don't use it anymore

bolshakov:
> However, I took another look on how Configuration object is implemented and it looks like we can do better here.
Currently, if we use the new interface, we actually use Stoplight::Configuration, which either handles method calls himself or delegates those method calls to Light object. I think that it's too much responsibility for the Configuration object.

@Lokideos We can try to split configuration into configuration and builder, but the `Light` currently mixes and matches these features anyway, and I don't want to break this right now. Please take a look at the recent changes.

bolshakov:
> Finally, we might want to write some examples of usage for our users using this new syntax.

I agree with that, but I want to agree on the interface before changing documentation.


> At least one example for Rails (e.g., we can create all needed configuration in a separate Stoplight initializer and use it later).
> WDYT?

I don't follow. Would you give me an example?

lib/stoplight.rb
```diff
@@ -1,6 +1,19 @@
...
+  class << self
+    # @!attribute default_data_store
+    #   @return [DataStore::Base]
+    attr_accessor :default_data_store
+
+    # @!attribute default_notifiers
+    #   @return [Array<Notifier::Base>]
+    attr_accessor :default_notifiers
+
+    # @!attribute default_notifiers
```
bolshakov, line 13:
```suggestion
    # @!attribute default_error_notifier
```

Lokideos (approved):

PR #261, opened 2024-12-30
Add link for Honeybadger Notifiers gem

ckornaros:
We use Honeybadger and Stoplight. We will migrate to the V4 soon so we decide to develop the gem to be able to notify HB. as it's not support anymore with the new version.


bolshakov:
Thank you for you contribution ❤️

PR #265, opened 2025-02-07
Ensure array is returned when cleaning failures in memory

barthez:
In memory data storage, when cleaning failures when the light name key is missing we returned nil instead of empty array. This caused errors in `on_success` callback checking `.empty?` on failures list.

### DIFF
--- a/lib/stoplight/data_store/memory.rb
+++ b/lib/stoplight/data_store/memory.rb
@@ -42,7 +42,7 @@ module Stoplight
       end
 
       def clear_failures(light)
-        synchronize { @failures.delete(light.name) }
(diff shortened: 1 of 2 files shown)


bolshakov:
Thank you for your contribution, Bartek. I'll release the fix shortly

bolshakov (approved):

bolshakov:
@barthez, for some reason, github does  not show your PR merged, but I've indeed merged it (see https://github.com/bolshakov/stoplight/commits/master/) I released Stoplight [v4.1.1](https://rubygems.org/gems/stoplight/versions/4.1.1) with your fix.

PR #274, opened 2025-04-05
Improving Configuration and Error Handling

bolshakov:
This PR introduces significant improvements to the Stoplight gem API, making it more flexible and easier to use. Note that this is a breaking change that addresses design limitations in the current implementation.

### Stateful Configuration Issues

The current approach using `#with_error_handler` and `#with_fallback` creates several problems:

1. **Configuration Leakage**: Both `#with_error_handler` and `#with_fallback` were slightly different from other configuration options, it was possible to define them on the `Stoplight::Light` lever which was a mutable instance. It was implemented this way for the sake of backward compatibility.

   ```ruby
   # This light is used across different services
   shared_light = Stoplight("payment")
     .with_threshold(5)
     .with_fallback { |error| log_and_return_empty_array(error) }
   
   # Later in code path A - modifies the global light object
   shared_light.with_error_handler { |error, handler| .... l }.run { ... }
   
   # Later in code path B - now has the error handler from code path A!
   shared_light.run { fetch_data } # Uses the fallback from path A if it fails
   ```

2. **Block-based Chain Breaking**: The block-based API breaks the natural method chaining:

   ```ruby
   # Awkward nesting required when combining multiple blocks
   Stoplight("example").with_fallback do |e| 
      handle(e) 
   end.run { code }
   ```

### Error Handler Complexity

The current `#with_error_handler` design is confusing:

```ruby
light.with_error_handler do |error, handle|
  if error.is_a?(ActiveRecord::RecordNotFound)
    raise error  # Skip this error
  else      
    handle.call(error)  # Default behavior
  end
end
```

This pattern:
- Requires understanding a complex behavioral contract
- Creates confusing control flow where raising an error is the mechanism to skip it
- Makes it easy to introduce bugs by forgetting to call the handler

### 1. Runtime Fallbacks

Fallbacks are now provided directly to the `#run` method as a parameter:

```ruby
light = Stoplight("payment-gateway")

# Different operations with operation-specific fallbacks
light.run(-> { [] }) { get_invoices }
light.run(-> { 0 }) { get_credits }
```

Benefits:
- **Operation-specific fallbacks**: Each protected operation can have its own appropriate fallback
- **Shared configuration**: Core circuit breaker settings can be reused without fallback contamination
- **Clear intent**: Makes it obvious which fallback belongs to which operation

### 2. Simplified Error Handling

Replaced the complex `#with_error_handler` approach with two declarative methods:

```ruby
light = Stoplight('api-service')
  .with_tracked_errors(NetworkError, Timeout::Error)
  .with_skipped_errors(ActiveRecord::RecordNotFound)
```

Benefits:
- **Declarative over imperative**: Simply list error types instead of writing custom handler logic
- **Easier to understand**: Clear separation between "count this error" and "ignore this error"
- **Reduced chance of bugs**: No possibility of forgetting to call the handler or raise the error

## Migration Path

Since this is a breaking change, existing code will need updates:

```ruby
# Old code
light = Stoplight('example')
  .with_error_handler do |error, handle|
    if error.is_a?(ActiveRecord::RecordNotFound)
      raise error
    else      
      handle.call(error)
    end
  end
  .with_fallback { |error| [] }

# New code
light = Stoplight('example')
  .with_skipped_errors(ActiveRecord::RecordNotFound)

light.run(->() { [] }) { code_to_protect }
```

We recommend updating in stages:
1. First migrate error handlers to the new pattern
2. Then update run calls to include fallbacks

## Why This Breaking Change is Necessary

The limitations of the current design cannot be addressed without breaking changes because:

1. The mutable, stateful nature of lights is fundamental to the current implementation
2. The block-based API for error handling and fallbacks can't be cleanly replaced without a new interface
3. The improved approach simplifies code, reduces bugs, and enables important use cases that were impossible before

While we recognize this is a breaking change, the significantly improved API and resolved design problems justify the transition. Getting rid of mutable state from `Stoplight::Light` instances allows us to perform massive refactoring unifying several internal entities (Bulder, Light, CircuitBreaker) into one. This simplifies the code and therefore maintenance

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,3 +25,6 @@ Layout/BlockAlignment:
 Metrics/ParameterLists:
   Exclude:
     - 'lib/stoplight/configuration.rb'
+
+Style/CaseEquality:
+  Enabled: false
--- a/lib/stoplight/circuit_breaker.rb
+++ b/lib/stoplight/circuit_breaker.rb
@@ -104,19 +104,6 @@ module Stoplight
       raise NotImplementedError
     end
 
-    # Configures light with the given fallback block
-    #
-    # @example
-    #   light = Stoplight('example')
-    #   light.with_fallback { |error| e.is_a?()ZeroDivisionError) ? 0 : nil }
-    #   light.run { 1 / 0} #=> 0
-    #
-    # @yieldparam error [Exception, nil]
-    # @return [Stoplight::CircuitBreaker]
-    def with_fallback(&fallback)
-      raise NotImplementedError
-    end
-
     # @return [String] one of +locked_green+, +locked_red+, and +unlocked+
     def state
       raise NotImplementedError
@@ -147,9 +134,14 @@ module Stoplight
     #   light = Stoplight('example')
     #   light.run { 2/0 }
     #
+    # @example Running with fallback
+    #   light = Stoplight('example')
+    #   light.run(->(error) { 0 }) { 1 / 0 } #=> 0
+    #
+    # @param fallback [Proc, nil] (nil) fallback code to run if the circuit breaker is open
     # @raise [Stoplight::Error::RedLight]
     # @return [any]
-    def run(&code)
+    def run(fallback = nil, &code)
       raise NotImplementedError
     end
 
--- a/lib/stoplight/circuit_breaker.rb
+++ b/lib/stoplight/circuit_breaker.rb
@@ -83,6 +83,45 @@ module Stoplight
       reconfigure(configuration.with(error_notifier: error_notifier))
     end
 
+    # Configures a custom list of tracked errors that counts toward the threshold.
+    #
+    # @example
+    #   light = Stoplight('example')
+    #     .with_tracked_errors(TimeoutError, NetworkError)
+    #   light.run { call_external_service }
+    #
+    # In the example above, the +TimeoutError+ and +NetworkError+ exceptions
+    # will be counted towards the threshold for moving the circuit breaker into the red state.
+    # If not configured, the default tracked error is +StandardError+.
+    #
+    # @param tracked_errors [Array<StandardError>]
+    # @return [Stoplight::CircuitBreaker]
+    def with_tracked_errors(*tracked_errors)
+      reconfigure(configuration.with(tracked_errors: tracked_errors.dup.freeze))
+    end
+
+    # Configures a custom list of skipped errors that do not count toward the threshold.
+    # Typically, such errors does not represent a real failure and handled somewhere else
+    # in the code.
+    #
+    # @example
+    #   light = Stoplight('example')
+    #    .with_skipped_errors(ActiveRecord::RecordNotFound)
+    #   light.run { User.find(123) }
+    #
+    # In the example above, the +ActiveRecord::RecordNotFound+ doesn't
+    # move the circuit breaker into the red state.
+    #
+    # The list of skipped errors is always complemented by the default
+    # skipped errors: +NoMemoryError+, +ScriptError+, +SecurityError+, etc.
+    # @see +Stoplight::Default::SKIPPED_ERRORS+
+    #
+    # @param skipped_errors [Array<Exception>]
+    # @return [Stoplight::CircuitBreaker]
+    def with_skipped_errors(*skipped_errors)
+      reconfigure(configuration.with(skipped_errors: skipped_errors))
+    end
+
     # Configures a custom proc that allows you not to handle an error
     # with Stoplight.
     #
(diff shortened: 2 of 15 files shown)


lib/stoplight/error.rb
```diff
@@ -2,10 +2,6 @@
 
 module Stoplight
   module Error
-    HANDLER = lambda do |error|
-      raise error if AVOID_RESCUING.any? { |klass| error.is_a?(klass) }
-    end
-
     AVOID_RESCUING = [
```
Lokideos, line 5: `NIT` Do we still use `AVOID_RESCUING` after removing the `HANDLER`?

lib/stoplight/light/runnable.rb
Lokideos, line 1: `NIT` I still think that we should do something with `Light::Runnable#color` method in future refactorings 🙂

lib/stoplight/light/runnable.rb
```diff
@@ -24,63 +24,71 @@ def color
...
+      def run_red(fallback)
         raise Error::RedLight, name unless fallback
 
         fallback.call(nil)
       end
 
-      def run_code(on_success, on_failure, &code)
+      def run_code(on_success, on_failure, fallback, &code)
         result = code.call
         failures = clear_failures
```
Lokideos, line 68: It looks like we do not check for the time window if we use in-memory storage, so there might be a bug here, but I'm not sure about this. I'll take a closer look at it tomorrow 🙂

Lokideos (approved):
Amazing job! 🚀 
I'm not sure if `#run_code` works correctly for InMemory storage, though 🤔 But it is not related to the changes in the PR anyway 🙂

.rubocop.yml
```diff
@@ -25,3 +25,6 @@ Layout/BlockAlignment:
 Metrics/ParameterLists:
   Exclude:
     - 'lib/stoplight/configuration.rb'
+
+Style/CaseEquality:
```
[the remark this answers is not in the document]
bolshakov: Not this time. I'm not sure why rubocop does not like `===` but it's usage here is a design choice. It allows passing error classes and procs as errors to the tracked errors and skipped errors lists. Similar to how you can pass anything to case/when and compare using `===`.

lib/stoplight/circuit_breaker.rb
```diff
@@ -147,9 +152,14 @@ def color
...
     #
+    # @example Running with fallback
+    #   light = Stoplight('example')
+    #   light.run(->(error) { 0 }) { 1 / 0 } #=> 0
+    #
+    # @param fallback [Proc, nil] (nil) fallback code to run if the circuit breaker is open
     # @raise [Stoplight::Error::RedLight]
     # @return [any]
-    def run(&code)
+    def run(fallback = nil, &code)
```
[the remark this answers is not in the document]
bolshakov: I thought about this a designed in favor of positional argument here, because the method signature is quite symilar to stdlib's `find`: 

```ruby
list.find(-> { defaul }) { |x| x == 1 }
```

To be honest not sure if it matters. Does this parity with find makes sense in you opinion?

lib/stoplight/configuration.rb
```diff
@@ -95,16 +112,14 @@ def with(
       name: self.name,
       notifiers: self.notifiers,
       threshold: self.threshold,
-      window_size: self.window_size
+      window_size: self.window_size,
+      tracked_errors: self.tracked_errors,
+      skipped_errors: self.skipped_errors
     )
       Configuration.new(
```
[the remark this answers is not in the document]
bolshakov: Totally agree, but rubocop thinks I the less linkes is better. TBH, I'm sick of rubocop and thinking of replacing it with [standard](https://github.com/standardrb/standard) which is less opinionated and has more senseble defaults.

lib/stoplight/error.rb
bolshakov: Good catch

lib/stoplight/light/runnable.rb
bolshakov: yeah, it's quite meaty

lib/stoplight/light/runnable.rb
bolshakov: Thanks a bunch 🙇

lib/stoplight/circuit_breaker.rb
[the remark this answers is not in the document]
Lokideos: > To be honest not sure if it matters. Does this parity with find makes sense in you opinion?

I think we can leave it as is, especially for one argument.

lib/stoplight/configuration.rb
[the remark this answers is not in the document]
Lokideos: Never heard about this gem, but I'm up for trying it out. Looks promising 🙂

PR #277, opened 2025-04-06
Refactor Stoplight internal architecture

### LINKED ISSUE #278: Drop deprecated interfaces from Stoplight
Drop `Stoplight::Builder` that was temporarily introduced to simplify migration from Stoplight 3.x to Stoplight 4.x

bolshakov:
Closes #278

This PR refactors the internal architecture of Stoplight to simplify its design and improve maintainability. The primary goal was to eliminate unnecessary indirection by replacing the Builder pattern with a more direct approach while maintaining backward compatibility.

## Changes
- Replaced `Stoplight::Builder` with a more straightforward implementation that directly uses `Stoplight::Light`
- Renamed `Stoplight::Configuration` to `Stoplight::Light::Config` to better reflect its relationship with `Light`
- Extracted configuration-related functionality into a separate `Stoplight::Light::Configurable` module (similar to `Stoplight::Light::Runnable`, and `Stoplight::Light::Lockable`
- Simplified `Stoplight::CircuitBreaker` from a module to a class for backward compatibility
- Updated all data store implementations to work with the new configuration approach
- Reorganized the documentation of custom fallback in README for better flow

## Benefits
- More intuitive codebase structure with less indirection
- Reduced complexity by eliminating unnecessary abstraction layers
- Better separation of concerns through module extraction
- The public API remains backward compatible

## API Changes
- No breaking changes to the public API
- All existing code that uses Stoplight should continue to work without modification
- Internal implementation details have changed significantly but are hidden from consumers


lib/stoplight/light/lockable.rb
```diff
@@ -20,23 +20,36 @@ class Light
...
                 end
 
-        safely { data_store.set_state(self, state) }
+        safely { config.data_store.set_state(config, state) }
 
         self
       end
 
-      # @return [Stoplight::Light] returns unlocked light
+      # Unlocks light and sets it's state to State::UNLOCKED
```
bolshakov, line 43:
```suggestion
      # Unlocks light and sets its state to State::UNLOCKED
```

README.md
```diff
@@ -322,11 +320,7 @@ class ApplicationController < ActionController::Base
 
   def stoplight(&block)
     Stoplight("#{params[:controller]}##{params[:action]}")
-      .with_fallback do |error|
-        Rails.logger.error(error)
-        render(nothing: true, status: :service_unavailable)
-      end
-      .run(&block)
+      .run(-> { render(nothing: true, status: :service_unavailable) }, &block)
```
Lokideos, line 323: Should we keep the fallback (but with the new implementation) in this example?

lib/stoplight/light.rb
```diff
@@ -3,55 +3,25 @@
...
     end
 
-    private
-
-    def reconfigure(configuration)
-      @configuration = configuration
-      self
+    # @param other [any]
+    # @return [Boolean]
+    def ==(other)
```
Lokideos, line 23: Where do we use this public method? 🤔

spec/support/light/runnable/state.rb
```diff
@@ -1,18 +1,15 @@
 # frozen_string_literal: true
 
 RSpec.shared_examples 'Stoplight::Light::Runnable#state' do
-  subject(:light) { Stoplight(name).build }
-
   let(:name) { random_string }
-  let(:data_store) { light.configuration.data_store }
```
Lokideos, line 7: Now we rely on defaults, right?

Lokideos (approved):
🚀

README.md
bolshakov: Honestly, I don't think calling logger from fallback function is an intedended fallback use considering Stoplight offers configurable errors notifiers. Another issues with the original example is that when Stoplight is red, it will pass `nil` to the follback function which leads to calling `Ralis.logger.error(nil)`. This is not a new beheviour, it was implemented this way before. 

So I wanted to update this example to showcase a more ideomatic use case. Does it make sense?

lib/stoplight/light.rb
bolshakov: I don't think we use this method within the Stoplight itself. The idea behind this method is to enable users to compare stoplights, for instance, during tests, and consider two lights with the same configuration to be equivalent. 

- [ ] Test `Stoplight::Light#==`

PR #284, opened 2025-04-09
Drop Redlock gem in favor of notification deduplication mechanism

### LINKED ISSUE #276: Update RedLock version
https://github.com/bolshakov/stoplight/pull/225

RedLock version is a long time outdated. We need to update its version and release it as a minor Stoplight version without bringing backward incompatible changes from the master branch

bolshakov:
Closes #276

- Renamed `with_notification_lock` to `with_deduplicated_notification` to accurately reflect its purpose
- Implemented a Redis-based notification deduplication system using Lua scripting

## Why
The previous approach attempted to use a locking mechanism, but what we actually needed was notification deduplication. This change:
- Eliminates unnecessary complexity and dependencies
- More precisely addresses the actual problem: preventing duplicate notifications
- Provides better performance by using atomic Lua operations

## Breaking changes
This PR introduces a backward incompatible change to the `Stoplight::DataStore::Base` interface:
`with_notification_lock` was renamed to `with_deduplicated_notification`

## Who is affected
- Developers who have created custom data store implementations that inherit from `Stoplight::DataStore::Base`
- Users who directly call the `with_notification_lock` method

## Migration path
Custom data store implementations should rename their method from `with_notification_lock` to `with_deduplicated_notification` and 
update implementation to match the new signature.


lib/stoplight/data_store/redis.rb
```diff
@@ -94,16 +92,50 @@ def clear_state(config)
...
+
+        if prev_from_color == from_color and prev_to_color == to_color then
+          return 0
+        else
+          redis.call('HSET', last_notification_key, 'from_color', from_color, 'to_color', to_color)
+          redis.call('EXPIRE', last_notification_key, ttl)
+          return 1
         end
+      LUA
+      private_constant :NOTIFICATION_DEDUPLICATION_SCRIPT
```
Lokideos, line 129: Brilliant 👏

lib/stoplight/data_store/redis.rb
```diff
@@ -94,16 +92,50 @@ def clear_state(config)
...
+      # 3. If it's a different transition (or first time), it:
+      #    - Records the new transition with a timestamp
+      #    - Sets an expiration time on the record (allowing notifications to repeat after TTL as a fail-safe mechanism)
+      #    - Returns 1 (proceed with notification)
+      #
+      # This ensures that when multiple servers detect the same transition simultaneously,
+      # only one notification is sent, while still allowing future transitions to trigger notifications.
+      #
+      # This script is executed on redis and it's guaranteed that the operations are atomic.
+      NOTIFICATION_DEDUPLICATION_SCRIPT = <<~LUA
```
Lokideos, line 110: If we use more LUA scripts in the future, I'd place them separately from the `data_store/redis.rb` file (e.g., `data_store/lua_scripts` or something).

lib/stoplight/data_store/memory.rb
```diff
@@ -57,7 +57,7 @@ def clear_state(config)
         synchronize { @states.delete(config.name) }
       end
 
-      def with_notification_lock(config, from_color, to_color)
+      def with_deduplicated_notification(config, from_color, to_color)
```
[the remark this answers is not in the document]
Lokideos: Yep, it's obvious, though, since we use in-memory storage.

Lokideos (approved):
Great work 🚀 
Speaking of notifications, I think we should update notifiers (or at least [sentry notifier](https://github.com/bolshakov/stoplight-sentry)) once we have time 🤔 
Since it lists already incorrect syntax to add a notifier in its readme 🙂

bolshakov:
@Lokideos good catch, I created a ticket not get this lost https://github.com/bolshakov/stoplight/issues/286#issue-2984723729

lib/stoplight/data_store/redis.rb
bolshakov: This is a good idea. We can use the Lua script to store errors as well. Let's keep it as is for now, but this is definitely a way to go if we have more such scripts

PR #288, opened 2025-04-10
Replace RuboCop with Standard gem for code linting

bolshakov:
* Standard provides a zero-configuration approach, reducing time spent on style debates
* Eliminates the need to maintain custom `.rubocop.yml` configuration files
* Based on RuboCop but with opinionated defaults aligned with community practices (and to be honest with my personal preferences)


lib/stoplight/data_store/memory.rb
```diff
@@ -1,19 +1,19 @@
...
     class Memory < Base
       include MonitorMixin
-      KEY_SEPARATOR = ':'
+      KEY_SEPARATOR = ":"
 
       def initialize
         @failures = Hash.new { |h, k| h[k] = [] }
         @states = Hash.new { |h, k| h[k] = State::UNLOCKED }
         @last_notifications = {}
-        super() # MonitorMixin
```
Lokideos, line 16: Is this a linter fix, or do we no longer want to depend on the parent method arguments?

lib/stoplight/data_store/redis.rb
```diff
@@ -160,7 +160,7 @@ def query_failures(config, transaction: @redis)
       def normalize_failures(failures, error_notifier)
         failures.map do |json|
           Failure.from_json(json)
-        rescue StandardError => e
```
Lokideos, line 163: I mean, I prefer explicit errors, but that works 🙂

Lokideos (approved):
🚀

lib/stoplight/data_store/memory.rb
bolshakov: Yep this is `bundle exec standardrb --fix` 

> Style/SuperArguments: Call `super` without arguments and parentheses when the signature is identical.

As any opinionated tool, it may have opinions that we disagree with, but in general, I find its default settings more sensible.

lib/stoplight/data_store/redis.rb
bolshakov: This is an interesting one. I can easily defend both approaches, but the beauty of the “standard” gem is that, in general, it has fewer opinions on stylistic matters. I don’t quite understand why Rubocop wants you to assign any exception variable, `e`, but the “standard” gem is fine with `e` as well as with `error`. Another notable difference is that the “standard” gem doesn’t force you to hang long lines in the middle of the screen, but if you prefer, you can do that. In that sense, the “standard” gem doesn’t have any opinion on it. This is the price I'm ready to pay to stress less about things that does not metter that much

lib/stoplight/data_store/memory.rb
Lokideos: Oh, I thought that method signatures were not identical. I'm good with this change then.

Lokideos:
@bolshakov  Yep, I'm good. 🙂

Issue #285, opened 2025-04-10
Performance Optimizations


bolshakov:
To improve performance, we would like to explore potential bottlenecks

<details>
  <summary>Profiling results</summary>

```ruby
# frozen_string_literal: true

require "ruby-prof"
require "stoplight"
require "redis"
require "fileutils"

# Create output directory
FileUtils.mkdir_p("profile_results")

# Setup Stoplight with Redis data store
redis = Redis.new
data_store = Stoplight::DataStore::Redis.new(redis)
Stoplight.default_data_store = data_store

stoplight = Stoplight("example").with_threshold(5).with_cool_off_time(60)

def profile_scenario(name, measurement_mode = RubyProf::WALL_TIME)
  # Start profiling with specified mode
  RubyProf.measure_mode = measurement_mode
  RubyProf.start

  yield

  result = RubyProf.stop

  File.open("profile_results/#{name}_flat.txt", "w") do |file|
    RubyProf::FlatPrinter.new(result).print(file)
  end

  File.open("profile_results/#{name}_graph.html", "w") do |file|
    RubyProf::GraphHtmlPrinter.new(result).print(file)
  end

  File.open("profile_results/#{name}_callstack.html", "w") do |file|
    RubyProf::CallStackPrinter.new(result).print(file)
  end
end

profile_scenario("success", RubyProf::PROCESS_TIME) do
  50.times do
    stoplight.run do
      sleep(0.01)
      10_000.times { Math.sqrt(rand) }
      File.read("/dev/null")
    end
  end
end

puts "Profiling complete. Results saved in profile_results/ directory."
```

  ```pre
Measure Mode: process_time
Thread ID: 2320
Fiber ID: 2300
Total: 2.073490
Sort by: self_time

 %self      total      self      wait     child     calls  name                           location
 60.12      2.073     1.247     0.000     0.827       51  *Integer#times                  <internal:numeric>:231
  8.96      0.186     0.186     0.000     0.000   500000   <Module::Math>#sqrt            
  8.89      0.184     0.184     0.000     0.000   500000   Kernel#rand                    
  8.85      0.183     0.183     0.000     0.000   500351   Integer#<                      
  8.79      0.182     0.182     0.000     0.000   500050   Integer#succ                   
  0.55      0.055     0.011     0.000     0.043     1501  *Array#each                     
  0.18      0.011     0.004     0.000     0.007      576   Kernel#loop                    <internal:kernel>:180
[...]

PR #293, opened 2025-04-12
Optimize notification locking for improved responsiveness

bolshakov:
The original implementation of `#with_deduplicated_notification` in `Memory` data store was locking the entire data store via MonitorMixin's `synchronize` while yielding to notifiers. Even though notifications are rare in our system, when they do happen, they could potentially block other critical operations.

This change introduces a focused optimization:

1. Separate the notification lock from the main data store lock by using a dedicated `Thread::Mutex` specifically for notifications
2. Move the yield operation outside of the synchronized block to prevent blocking other operations during notification processing
3. Add benchmarks to measure and verify the performance characteristics

Under simulated high contention (10 resource names, 16 threads with 0.001s operations):
- Original implementation: ~36.7s real time
- Optimized implementation: ~2.8s real time (13x faster)

Even under low contention scenarios, which better represent our production environment, the optimized version showed similar improvements in responsiveness.

While notifications are infrequent in our system, the impact of blocking the entire data store during a notification can still affect system responsiveness. This change ensures that when notifications do occur, they don't interfere with other critical operations that might be happening simultaneously.

The optimization maintains the same semantics and thread safety guarantees while significantly reducing the potential for lock contention during these occasional but important notification events.

#285

### DIFF
--- a/lib/stoplight/data_store/memory.rb
+++ b/lib/stoplight/data_store/memory.rb
@@ -12,6 +12,7 @@ module Stoplight
       def initialize
         @failures = Hash.new { |h, k| h[k] = [] }
         @states = Hash.new { |h, k| h[k] = State::UNLOCKED }
+        @notification_lock = Monitor.new
         @last_notifications = {}
         super # MonitorMixin
       end
@@ -58,13 +59,14 @@ module Stoplight
       end
 
       def with_deduplicated_notification(config, from_color, to_color)
-        synchronize do
+        notify = false
+        @notification_lock.synchronize do
           if last_notification(config) != [from_color, to_color]
             set_last_notification(config, from_color, to_color)
-
-            yield
+            notify = true
           end
         end
+        yield if notify
       end
 
       private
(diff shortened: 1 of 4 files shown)


bench/memory/with_deduplicated_notification.rb
Lokideos, line 1: Maybe we should rename the folder to `memory_data_store` or something like that to stress that we actually check for time elapsed and do not try to profile memory usage? 🤔

lib/stoplight/data_store/memory.rb
Lokideos, line 1: 🔥

Lokideos (approved):
Great work as always 🙂

bench/memory/with_deduplicated_notification.rb
bolshakov: Right, makes sense! I'd like to keep these benchmarks in the repo so we can revise them from time to time. I noticed that some optimizations tend to be less important with new ruby versions while complicating the code. Not in this specific case, of course

PR #301, opened 2025-04-15
Use zeitwerk for code loading

bolshakov:
It was always a pain to maintain the right order of components loading. With zeitwerk, we don't have to worry about this anymore. It loads the code automatically.

### DIFF
--- a/lib/stoplight.rb
+++ b/lib/stoplight.rb
@@ -1,5 +1,12 @@
 # frozen_string_literal: true
 
+require "zeitwerk"
+
+loader = Zeitwerk::Loader.for_gem
+loader.inflector.inflect("io" => "IO")
+loader.ignore("#{__dir__}/stoplight/rspec.rb", "#{__dir__}/stoplight/rspec")
+loader.setup
+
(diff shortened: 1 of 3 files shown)


lib/stoplight.rb
```diff
@@ -1,5 +1,12 @@
 # frozen_string_literal: true
 
+require "zeitwerk"
+
+loader = Zeitwerk::Loader.for_gem
+loader.inflector.inflect("io" => "IO")
+loader.ignore("#{__dir__}/stoplight/rspec.rb", "#{__dir__}/stoplight/rspec")
```
Lokideos, line 7: Should we also ignore `bench/**`?

Lokideos (approved):
🚀

lib/stoplight.rb
bolshakov: It’s out of the `lib` directory so ignored by default. If it was not ignored, it would raise a Zeitwerk error -- Bench constant not found or something like this. 

Just to be cleal, here ignore shared RSpec helpers for user-defined notifiers, not specs for Stoplight itself.

PR #307, opened 2025-04-21
Refactor stoplight using Strategy pattern for improved maintainability

bolshakov:
This refactors the `Runnable` module implementation by introducing the Strategy pattern for handling different light states (Green, Yellow, Red). The existing fail-safe (`Runnable#safely`) mechanisms are enhanced through dedicated wrapper classes, resulting in a more maintainable and better-organized codebase.

The current implementation has complex logic inside the Runnable module:

* The run method contains nested conditional logic for each light state
* Error handling and fail-safe behaviors are mixed with business logic
* The safely method provides error resilience but is coupled with the main implementation

This refactoring improves maintainability by separating these concerns and making the code easier to understand and extend. 

## Performance Impact 

This change delivers significant performance improvement in some scenarios. Below you can find detailed benchmarks:

## Call of pre-created light:

<details>
<summary>Memory Data Store Benchmark</summary>

```ruby
require "benchmark/ips"
require "stoplight"
cashed_stoplight = Stoplight("")

Benchmark.ips do |b|
  b.report("Call Before Refactoring") { cashed_stoplight.run {} }
  b.report("Call After Refactoring") { cashed_stoplight.run {} }

  b.hold! "compare 3"

  b.compare!
end
```

</details>

<details>
<summary>Redis Data Store Benchmark</summary>

```ruby
require "benchmark/ips"
require "stoplight"
require "redis"

redis = Redis.new
data_store = Stoplight::DataStore::Redis.new(redis)
  config.data_store = data_store
end
cashed_stoplight = Stoplight("")

Benchmark.ips do |b|
  b.report("Call Before Refactoring") { cashed_stoplight.run {} }
  b.report("Call After Refactoring") { cashed_stoplight.run {} }

  b.hold! "compare 1"

  b.compare!
end
```
</details>

In memory data store appears to be 35% faster after the refactoring:

```
Comparison:
Call After Refactoring:   931387.5 i/s
Call Before Refactoring:   691296.5 i/s - 1.35x  slower
```

Redis data store shows no statistically significant difference:

```
Comparison:
Call After Refactoring:     9823.7 i/s
Call Before Refactoring:     9172.6 i/s - same-ish: difference falls within error
```

## Create & Call a light

In this bencmark I tested the performance of creating and immediately calling stoplight:

<details>
<summary>Memory Data Store Benchmark</summary>

```ruby
Benchmark.ips do |b|
  b.report("Create & Call Before Refactoring") { Stoplight("").run {} }
  b.report("Create & Call After Refactoring") { Stoplight("").run {} }

  b.hold! "compare 4"

  b.compare!
end
```

</details>

<details>
<summary>Redis Data Store Benchmark</summary>

```ruby
# frozen_string_literal: true

require "benchmark/ips"
require "stoplight"
require "redis"

redis = Redis.new
data_store = Stoplight::DataStore::Redis.new(redis)
  config.data_store = data_store
end

Benchmark.ips do |b|
  b.report("Create & Call Before Refactoring") { Stoplight("").run {} }
  b.report("Create & Call After Refactoring") { Stoplight("").run {} }

  b.hold! "compare 2"

  b.compare!
end

```

</details>

Redis data store shows no statistically significant difference:

```
Comparison:
Create & Call After Refactoring:   193201.5 i/s
Create & Call Before Refactoring:   193079.6 i/s - same-ish: difference falls within error
```

Redis data store  appears to be 21% faster after the refactoring:

```
Comparison:
Create & Call After Refactoring:    10022.7 i/s
Create & Call Before Refactoring:     8299.5 i/s - 1.21x  slower
```

#285


UPGRADING.md
```diff
@@ -33,14 +33,15 @@ Unfortunately, we cannot support all the possible notifiers.
...
+
 ```ruby
 # Gemfile
 gem 'sentry'
 gem 'stoplight'
 gem 'stoplight-sentry'
 
 # Code 
-Stoplight.default_notifiers += [Stoplight::Sentry::Notifier.new(Sentry)]
+Stoplight.default_notifiers += [Stoplight::Sentry::FailSafe.new(Sentry)]
```
bolshakov, line 44:
```suggestion
Stoplight.default_notifiers += [Stoplight::Sentry::Notifier.new(Sentry)]
```

lib/stoplight/light/config.rb
```diff
@@ -69,6 +69,27 @@ def ==(other)
...
+        skip = skipped_errors.any? { |klass| klass === error }
+        track = tracked_errors.any? { |klass| klass === error }
+
+        !skip && track
+      end
+
+      # @param number_of_errors [Numeric]
+      # @return [Boolean]
+      def threshold_exceeded?(number_of_errors)
+        number_of_errors == threshold
```
Lokideos, line 84:
```suggestion
        number_of_errors >= threshold
```
I'm not sure if we can arrive at this state (like if we fallback to memory and then go back to redis and find out that number of errors exceed the threshold), but I think we will be safer this way 🙂

lib/stoplight/light/runnable.rb
```diff
@@ -21,12 +19,12 @@ def state
       #
       # @return [String] returns current light color
       def color
-        failures, state = failures_and_state
+        failures, state = config.data_store.get_all(config)
```
Lokideos, line 22: I think we should hide `config.data_store.get_all(config)` behind some private method. WDYT?

lib/stoplight/light/runnable.rb
```diff
@@ -51,96 +49,20 @@ def color
...
-
-      def safely(default = nil, &code)
-        return yield if config.data_store == Default::DATA_STORE
-
-        fallback = proc do |error|
-          config.error_notifier.call(error) if error
-          default
+      private def state_strategy_factory(color)
+        @strategies ||= {}
+        @strategies[color] ||= case color
```
Lokideos, line 58:
```suggestion
        @strategies[color] = case color
```
I'm not sure if we want to memoize the light color here 🤔 What is your reasoning behind it?

lib/stoplight/light/runnable/green_run_strategy.rb
```diff
@@ -0,0 +1,49 @@
...
+        #
+        # @param fallback [Proc, nil] A fallback proc to execute in case of an error.
+        # @yield The code block to execute.
+        # @return [Object] The result of the code block if successful.
+        # @raise [Exception] Re-raises the error if it is not tracked or no fallback is provided.
+        def execute(fallback, &code)
+          code.call.tap do
+            data_store.clear_failures(config)
+          end
+        rescue Exception => error # rubocop: disable Lint/RescueException
```
Lokideos, line 23: Well, after all maybe `standard` isn't a good fit for `stoplight` 😞

lib/stoplight/light/runnable/run_strategy.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+module Stoplight
+  class Light < CircuitBreaker
+    module Runnable
+      # Represents a strategy for running a light's operations.
```
Lokideos, line 6:
```suggestion
      # Represents an abstract strategy for running a light's operations.
      # Every new strategy should be a child of this class.
```

lib/stoplight/light/runnable/yellow_run_strategy.rb
```diff
@@ -0,0 +1,37 @@
...
+        # Executes the provided code block when the light is in the yellow state.
+        #
+        # @param fallback [Proc, nil] A fallback proc to execute in case of an error.
+        # @yield The code block to execute.
+        # @return [Object] The result of the code block if successful.
+        # @raise [Exception] Re-raises the error if it is not tracked or no fallback is provided.
+        def execute(fallback, &code)
+          code.call.tap do
+            failures = data_store.clear_failures(config)
+            notify(config, Color::RED, Color::GREEN) unless failures.empty?
```
Lokideos, line 23:
```ruby
notify(config, Color::RED, Color::GREEN) unless failures.empty?
```
I'm not sure what the intention is here. We're trying to avoid duplicated notifications?

Lokideos (approved):
Overall great work as always 🚀 !
I see lines like `config.threshold_exceeded?`, and I think that we might miss an abstraction here (like `light.threshold_exceeded?` or `circuit_breaker.threshold_exceeded?`), but I might be wrong here.

lib/stoplight/light/runnable.rb
bolshakov: We're not. Here, we memoize only the value assigned to a specific color. This is a lazy initialization. I don't know why I did it in such a complex way. Let me rewrite it for clarity.

lib/stoplight/light/runnable.rb
bolshakov: I'm not sure, why?

lib/stoplight/light/runnable/yellow_run_strategy.rb
bolshakov: Yes, exactly. Similar to switching to the red state when we check the ex:act number of errors. I know this is confusing. Probably it worth explicitly mentioning this

PR #326, opened 2025-05-29
Remove CirciutBreaker class

bolshakov:


Lokideos (approved):
It's finally dead 💀 🎉 
If you never kill anything, you will live among zombies. And they will eat your brain. (c)

PR #328, opened 2025-05-29
Simplify state management methods

bolshakov:
Since all the metadata related to a Light is represented by a single unified Metadata object, we don't need a separate `get_state` method. The same information could be obtained by requesting the metadata itself. 

This PR also gets rid of `DataStore::Base#clear_state` since this method is not used inside the stoplight. For resetting the state, we offer the `Light#unlock` method. 

These cleanups make the public data store interface quite concise:

* `Stoplight::DataStore::Base#names` -- gets all the light names
* `Stoplight::DataStore::Base#get_metadata` queries metadata to a specific light
* `Stoplight::DataStore::Base#record_failure`
* `Stoplight::DataStore::Base#record_success`
* `Stoplight::DataStore::Base#record_recovery_probe_failure`
* `Stoplight::DataStore::Base#record_recovery_probe_success`
* `Stoplight::DataStore::Base#set_state`
* `Stoplight::DataStore::Base#transition_to_color`


lib/stoplight/light/runnable.rb
```diff
@@ -5,7 +5,10 @@ class Light < CircuitBreaker
     module Runnable # rubocop:disable Style/Documentation
       # @return [String]
       def state
-        config.data_store.get_state(config)
+        config
+          .data_store
+          .get_metadata(config)
+          .locked_state
```
Lokideos, line 11: I think it makes sense together with #327 👍

Lokideos (commented):
🚀

Lokideos (approved):

PR #338, opened 2025-06-06
Remove `Stoplight.default_*` configuration methods in favor of `Stoplght.configure` block

bolshakov:
This PR removes the `Stoplight.default_*` configuration methods introduced in `4.x` and replaces them with a new `Stoplight.configure` block pattern. This completes the configuration API evolution that began in `4.x` and provides a more robust, extensible configuration system for `5.x`.

**Removed methods:**

- `Stoplight.default_data_store=`
- `Stoplight.default_data_store`
- `Stoplight.default_notifiers=`
- `Stoplight.default_notifiers`
- `Stoplight.default_error_notifier=`
- `Stoplight.default_error_notifier`

```ruby
Stoplight.default_data_store = Redis.new
Stoplight.default_notifiers = [SlackNotifier.new]
Stoplight.default_error_notifier = ->(e) { Sentry.capture(e) }

  config.data_store = Redis.new
  config.notifiers = [SlackNotifier.new]
  config.error_notifier = ->(e) { Sentry.capture(e) }
end
```

* Rather than introducing another deprecation cycle that would require a `6.x` breaking change, we're removing all legacy configuration patterns in `5.x` to provide a clean foundation going forward
* This change aligns with other breaking changes in `5.x` (removal of deprecated code block patterns, module consolidation)
* The configure block provides atomic configuration and easier validation
* One comprehensive breaking change is better than multiple smaller ones across versions

- **Atomic**: All configuration happens in one place
- **Extensible**: Easy to add new configuration options
- **Validatable**: Can validate entire configuration at once
- **Thread-safe**: Single configuration point reduces race conditions

```ruby
Stoplight::Light.default_data_store = Redis.new

Stoplight.default_data_store = Redis.new

  config.data_store = Redis.new
end
```

This evolution reflects our commitment to finding the right long-term API design. Rather than perpetuating intermediate solutions with additional deprecation cycles, `5.x` establishes the configuration pattern we want to support going forward.

This change affects users who:
- Set global defaults using `Stoplight.default_*` methods
- Read current defaults using these methods in their code

**Note**: This is a **breaking change** appropriate for a major version release. While significant, this change eliminates multiple layers of legacy code and provides a clean, extensible foundation for future development. The migration path is straightforward and the new API is more robust than its predecessors.


Lokideos (approved):
Well, migration will be painful, but it has to be done sooner or later.

PR #343, opened 2025-06-16
feat: Generators

Lokideos:
This PR aims to add a simple generator to generate Stoplight initializer and Admin panel.
To generate Stoplight initializer run

```shell
rails g stoplight:install
```

To generate it with admin panel (by adding needed route to `config/routes.rb`) run:

```shell
rails g stoplight:install --with-admin-panel
```


Gemfile
```diff
@@ -28,4 +28,9 @@ group :development do
   gem "simplecov-lcov", "~> 0.8"
   gem "standard"
   gem "timecop", "~> 0.9"
+
+  # Rails dummy and ammeter dependencies
+  gem "activerecord"
+  gem "ammeter"
+  gem "pg"
```
bolshakov, line 35: Does it make sense to use the dummy app without active record to avoid having these deps at all?

Lokideos:
> noticed that the generated routes does not include authentication. Can we include some basic HTTP authentication by default? E.g. with `STOPLIGHT_ADMIN_USER` and `STOPLIGHT_ADMIN_PASSWORD` environment variables?

Sure! I'll do it.

.gitignore
```diff
@@ -4,7 +4,12 @@
...
 /pkg/
 /tmp/
+tmp/
 /vendor/
 /.idea
 profile_results
 /docker/Gemfile.lock
+
+spec/dummy/db/*.sqlite3
+spec/dummy/db/*.sqlite3-journal
```
bolshakov, line 14: not neaded anymore, right?

README.md
```diff
@@ -35,6 +35,14 @@ Or install it manually:
...
 ```
 
+### Rails
+
+For Rails application you can use generator to set up initializer and Admin Panel:
+
+```sh
+rails generate stoplight:install --with-admin-panel
+```
+
```
bolshakov, line 45: Hey! I think we should move this Rails generator section down to the bottom under a "Rails Integration" heading instead.

The thing that's bugging me is that we open the README by saying "Stoplight works with zero configuration out of the box," but then immediately after installation we're telling people to run a generator. That feels contradictory and might confuse new users into thinking they need to set up Redis and the admin panel just to get started.

I'd love to be more strategic about this positioning. The Redis data store and admin interface are really nice-to-have features for advanced users, but not everyone needs them (plenty of Rails apps don't even use Redis, right?).  What if we framed it more like "hey, if you're using Redis and want an admin interface, here's a handy generator that'll set everything up for you"?

That way we preserve that awesome zero-config promise upfront, and users who need the advanced features can discover them when they're ready. Does that make sense?

bolshakov (approved):
Oustanding work! 🔥 Please squash the commings before merging

PR #349, opened 2025-06-17
Feature/error rate strategy

bolshakov:


lib/stoplight/traffic_control/consecutive_failures.rb
```diff
@@ -14,18 +14,30 @@ module TrafficControl
...
+    #   traffic_control = Stoplight::TrafficControlStrategy::ConsecutiveFailures.new
+    #   config = Stoplight::Light::Config.new(threshold: 5, window_size: nil, traffic_control:)
     #
     # Will switch to red only if 5 consecutive failures occur regardless of the time window
     # @api private
     class ConsecutiveFailures < Base
+      # @param config [Stoplight::Light::Config]
+      # @return [Stoplight::Config::CompatibilityResult]
+      def check_compatibility(config)
+        if config.threshold < 1
```
Lokideos, line 32:
```suggestion
        if config.threshold <= 0
```
I think it's more clear 🤔 WDYT?

lib/stoplight/traffic_control/consecutive_failures.rb
```diff
@@ -14,18 +14,30 @@ module TrafficControl
...
-    #   strategy = Stoplight::TrafficControlStrategy::ConsecutiveFailures.new
+    #   traffic_control = Stoplight::TrafficControlStrategy::ConsecutiveFailures.new
+    #   config = Stoplight::Light::Config.new(threshold: 5, window_size: nil, traffic_control:)
     #
     # Will switch to red only if 5 consecutive failures occur regardless of the time window
     # @api private
     class ConsecutiveFailures < Base
+      # @param config [Stoplight::Light::Config]
+      # @return [Stoplight::Config::CompatibilityResult]
+      def check_compatibility(config)
```
Lokideos, line 31: Should we also check that `window_size` is either `nil` or positive `Integer` number?

Lokideos (approved):
###

lib/stoplight/traffic_control/consecutive_failures.rb
bolshakov: Yeah, changing the order of checks should make it more clear

PR #375, opened 2025-07-01
Add Configurable Consecutive Successes Recovery Strategy

bolshakov:
Replaces the `SingleSuccess` recovery strategy with a more robust `ConsecutiveSuccesses` strategy that requires a configurable number of successful recovery probes before resuming normal traffic flow.

The previous `SingleSuccess` strategy could cause circuit breaker flapping in scenarios where services intermittently fail during recovery. A single successful probe followed by failures would cause rapid state transitions between yellow and red, potentially overwhelming unstable services. 

## Configuration

```ruby
# Require 3 consecutive successful probes before resuming traffic
light = Stoplight("Payment API", recovery_threshold: 3)

# Default behavior (equivalent to previous SingleSuccess)
light = Stoplight("Payment API")  # recovery_threshold: 1
```

## Benefits

🔒 **Stable Recovery**: Prevents circuit breaker flapping by requiring proven stability over multiple probes instead of gambling on a single success. Services demonstrate consistent health before receiving full traffic.

⚡ **Still Fast**: Consecutive successes provide the fastest possible recovery once a service is truly stable. A genuinely healthy service will pass multiple probes quickly.

⚙️**Configurable Balance**: Teams can tune the stability vs. speed tradeoff per service - critical services use higher thresholds (3-5), while non-critical or lower-traffic services can use lower values (1-2) for faster recovery.

🔄 **Backward Compatible**: Existing deployments continue working unchanged with `recovery_threshold: 1`, maintaining current behavior.

This strategy provides the best of both worlds: preventing premature recovery that destabilizes systems while still allowing rapid recovery for truly healthy services.


README.md
```diff
@@ -331,6 +345,30 @@ Only evaluates error rate after at least 20 requests within the window. Default
 * **Consecutive Errors**: Low-medium traffic, simple behavior, occasional spikes expected
 * **Error Rate**: High traffic, percentage-based SLAs, variable traffic patterns
 
+### Traffic Recovery Strategies
+
+In yellow state, Stoplight behaves differently from normal operation. Instead of running the block of code, it
+first lets the underlying service cool down and then runs a test probe to see if the service has recovered.
+Traffic Recovery strategies control how Stoplight behaves during this recovery phase.
```
Lokideos, line 352: I think we should emphasize that during `Yellow` state the traffic flows freely, so even if we set `recovery_threshold` to 100, even the first request, if successfu,` will reach the target service and perform the required operation 🤔 
WDYT?

Lokideos (approved):
🚀 Stellar work!

README.md
bolshakov: Hey @Lokideos 👋 Thank you for the feedback. I rephrased this part and hope it now better conveys the actual behavior: 

```md
### Traffic Recovery Strategies

In the yellow state, Stoplight behaves differently from normal (green) operation. Instead of
blocking all traffic, it allows a limited number of real requests to pass through to
the underlying service to determine if it has recovered. These aren't synthetic probes -
they're actual user requests that will execute normally if the service is healthy.

After collecting the necessary data from these requests, Stoplight decides whether to
return to green or red state.

Traffic Recovery strategies control how Stoplight evaluates these requests during
the recovery phase.
```

What you think?

Lokideos: Sounds amazing! 🚀

PR #382, opened 2025-08-02
Fix notification about transition from yellow to red

### LINKED ISSUE #351: Better Traffic Recovery Strategies
The current `Stoplight::TrafficRecovery::SingleSuccess ` traffic recovery strategy presents significant reliability risks in production environments. The strategy implements a binary recovery approach where a single successful probe immediately resumes 100% traffic flow. This approach lacks nuance and fails to account for the complex nature of service recovery patterns, leading to service instability and cascade failures.

**Service Overwhelming**. Services that are partially recovered or still experiencing intermittent issues can be immediately overwhelmed when 100% of traffic resumes. 

**State Flapping**. Services experiencing intermittent failures cause rapid oscillation between RED and GREEN states. A service might succeed once, resume full traffic, fail again due to load, and immediately return to blocking all traffic. This creates an unpredictable user experience and can trigger cascade failures in dependent services.

## Proposed Solution Direction

* Stoplight should offer the ability to incorporate multiple success indicators rather than relying on a single probe result. This might include consecutive success requirements, success rate thresholds over time windows, or composite health metrics that consider latency and throughput alongside error rates.
* We should support incremental traffic increases over time, allowing services to warm up and preventing sudden load spikes. This requires implementing traffic sampling mechanisms and time-based progression. The question of whether this should be part of the recovery strategy configuration or the general stoplight configuration remains open.

## Implementation Considerations

The existing metadata structure provides a solid foundation for better recovery strategies. Available fields include:

* `recovery_probe_successes` and `recovery_probe_failures` for tracking probe results during recovery
* `consecutive_successes` and `consecutive_failures` for streak-based strategies
* `recovery_started_at` and `recovered_at` for time-based calculations
* `last_failure_at` and `last_success_at` for failure pattern analysis

Considering the rich metadata we have, we should strive to avoid significant infrastructure changes. Performance implications should be minimal since the required data tracking is already in place. The enhancement primarily involves leveraging existing metadata more intelligently rather than adding new Redis operations or complex state tracking.

At first, we want to offer advanced strategies keeping existing defaults.

bolshakov:
Closes #351

I implemented feature tests for traffic control recovery stategies and this revealed a bug in the `transition_to_yellow` implemenentation of the Memory data store:

```cucumber
  Background:
    Given a light "basic-service" exists with:
      | Recovery Threshold | 3 |
      | Cool Off Time     | 60 |
    And the protected service starts failing with "connection-timeout"
    And the light enters the red state
    And 60 seconds elapsed

  Scenario: Light returns to red after failure in yellow state
    Given the protected service recovers and starts functioning normally
    And I make 1 request to the protected service
    When the protected service starts failing with "connection-timeout" again
    Then the light color is yellow
    And I make 1 request to the protected service
    Then the light color is red
    And notification about transition from yellow to red is sent
```

This failed on the last step


Lokideos (approved):
🚀

Issue #416, opened 2025-08-23
Mention a way to run Stoplight with DragonflyDB


Lokideos:
Assess the work needed to support DragonflyDB in Stoplight to solve #414 
DragonflyDB utilizes a multi-threaded model, so that current Lua scripts will lead to race conditions.
We have to create a separate Data Store that guarantees transactional execution of Lua scripts with the DragonflyDB instance.

bolshakov:
Interesting, I've never heard of DragonflyDB, but looking at the [scripting documentation], it seems it's possible to configure it in a way incompatible with Redis Lua scripting. In particular, Redis Lua scripts are executed fully without commands of other clients in between, while Dragonfly's `disable-atomicity` flag makes Lua scripts behave more like a pipeline. 

The Redis data story relies on Redis's single-threaded architecture and Lua scripts' atomicity for serverless coordination. With Dragonfly, we seem to be able to use the same approach. 

I'm skeptical we can guarantee Redis data store is compatible with Dragonfly, but let's see if we can fix some low-hanging fruit to provide accidental compatibility (#415 is a good example) 



[scripting documentation] https://www.dragonflydb.io/docs/managing-dragonfly/scripting

Issue #414, opened 2025-08-22
`Redis::CommandError` when using Redis DataStore with DragonflyDB


ClearlyClaire:
**Describe the bug**

Since upgrading from Stoplight 4.4.1 to 5.3.1, we are regularly running into a `Redis::CommandError`:
`Switching stoplight:data_store:fail_safe:Stoplight::DataStore::Redis from green to red because Redis::CommandError ERR Error running script (call to 48d8273e290a4ccd1aacf5c24e828ddb25b88d5f): @user_script:15: attempt to compare number with string`

The script SHA seems to correspond to the `get_metadata.lua` script.

The error suggests that the value returned by https://github.com/bolshakov/stoplight/blob/3d45967b1667926cf100ce9b056b077fe875a926/lib/stoplight/data_store/redis/get_metadata.lua#L12 is a string, which might be a DragonflyDB quirk.

**To Reproduce**

We have found no clear way to reproduce this yet, the error occurs a lot in our logs, but we have not been able to manually reproducing it yet.

**Expected behavior**

The script runs without errors.

**Environment (please complete the following information):**
 - OS: [e.g. macOS 13.1 (22C65)]
 - Ruby [e.g. 3.0.4]
 - Version [e.g. 3.0.1]

**Additional context**

This seems to be the underlying issue in what we are encountering here: https://github.com/mastodon/mastodon/issues/35863

ClearlyClaire:
We have been able to reproduce the issue manually by having the Stoplight turn red, then green again:
```ruby
Stoplight('test-weird-stoplight-behavior', cool_off_time: 10, threshold: 2).run { raise ArgumentError } rescue ArgumentError
Stoplight('test-weird-stoplight-behavior', cool_off_time: 10, threshold: 2).run { raise ArgumentError } rescue ArgumentError
6.times do
  sleep 2
  Stoplight('test-weird-stoplight-behavior', cool_off_time: 10, threshold: 2).run { 'foo' } rescue Stoplight::Error::RedLight
end
```

Issue #413, opened 2025-08-22
`Stoplight::DataStore::Failsafe#record_failure` fallback return value causes `NoMethodError` in `Stoplight::Light::GreenRunStrategy#record_error`


ClearlyClaire:
**Describe the bug**

The fallback return value for `Stoplight::DataStore::Failsafe#record_failure` is `nil`, which causes `Stoplight::Light::GreenRunStrategy#record_error` to fail with:
```
NoMethodError: undefined 'consecutive_errors' for nil
```

**To Reproduce**

1. configure Stoplight with a Redis Datastore
2. run `Stoplight('bug-report-2025-08-22', cool_off_time: 60, threshold: 3).run { 'foo' }`
3. stop Redis
4. run `Stoplight('bug-report-2025-08-22', cool_off_time: 60, threshold: 3).run { raise 'foo' }`

**Expected behavior**

The raised exception is a `RuntimeError` with a `'foo'` message, or a `Stoplight::Error::RedLight` if it has reached the threshold.

**Additional context**

This is one part of the issues we are running into here: https://github.com/mastodon/mastodon/issues/35863

The underlying cause for the Redis Datastore sometimes failing is still under investigation.

PR #415, opened 2025-08-22
Fix `get_metadata.lua` script error on DragonflyDB

ClearlyClaire:
Fixes #414

PR #417, opened 2025-08-23
feat: Add DragonflyDB Support 

Lokideos:
This PR aims to add DragonflyDB support for Stoplight.

Addresses #413 #414 #416 

After some consideration, I've decided not to commit to support `DragonflyDB`. However, I will list recommendations regarding a way to run Stoplight alongside DragonflyDB, and we can accept contributions to support `DragonflyDB` in the future 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -377,7 +377,12 @@ single successful recovery probe will resume traffic flow.
 
 ### Data Store
 
-Stoplight uses an in-memory data store out of the box:
+Stoplight officially supports three data stores:
+- In-memory data store
+- Redis
+- Valkey
+
+By default, Stoplight uses an in-memory data store:
 
 ```ruby
 require "stoplight"
@@ -416,6 +421,12 @@ Stoplight.configure do |config|
   # ...
 end
 ```
+
+#### DragonflyDB Support
+
+Although Stoplight does not officially support [DragonflyDB], it can be used with it. For details, you may refer to the official [DragonflyDB documentation].
+
+**NOTE**: Compatibility with [DragonflyDB] is not guaranteed, and results may vary. However, you are welcome to contribute to the project if you find any issues.
 
 #### Connection Pooling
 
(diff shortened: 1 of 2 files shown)


bolshakov:
@Lokideos Thanks for taking the time to add DragonflyDB compatibility testing! I appreciate the effort and it's great to see the tests passing.

However, I'm hesitant to merge this as-is because it implies official support for DragonflyDB, which comes with some challenges:
- While DragonflyDB aims for Redis compatibility, it has different internals (multi-threaded architecture, different memory model)
- Circuit breaker logic relies heavily on precise timing and atomic operations via Lua scripts
- Subtle behavioral differences could surface in production scenarios that our test suite doesn't cover
- Supporting multiple backends exponentially increases the complexity of debugging issues
- Users would expect identical behavior across Redis and DragonflyDB
- Future DragonflyDB updates might introduce breaking changes we can't easily test against

**Additionally, I'm concerned about adding DragonflyDB to our main test suite because it could normalize having a red CI.** If DragonflyDB tests start failing due to compatibility issues or upstream changes, we risk creating alert fatigue and making it harder to spot real regressions in our core Redis support.

I'd be open to merging this if we frame it as "community-tested" rather than "officially supported":
1. Add documentation clearly stating Redis/Valkey is the only officially supported backend
2. Mark DragonflyDB tests as experimental (**and possibly run them separately from main CI** -- any ideas how exectly to make this distintion)
3. Include disclaimer that DragonflyDB support is community-contributed and may have edge cases. Welcome contributors to provide low-effort compatibility fixes such as #415

This would allow interested users to experiment while protecting the project from support obligations we can't fulfill.

What do you think about this approach?

README.md
```diff
@@ -417,6 +422,18 @@ Stoplight.configure do |config|
 end
 ```
 
+#### DragonflyDB Support
+
+Although Stoplight does not officially support [DragonflyDB], it can be used with it. To achieve behavior similar to Redis, start DragonflyDB with the following flags:
+
+```sh
+./dragonfly --default_lua_flags=allow-undeclared-keys,disable-atomicity
```
bolshakov, line 430: Doesn't it enforces a behavior the opposite to the Redis' default behavior?

Not sure about the behavior when the undeclared keys are allowed, but disabling atomicity makes Dragonfly's lua scripts behave more like Redis' pipelines - instructions from other clients can be executed between LUA Script's instructions making the begaviour completely unpredictable.

I suppose we should advise for bot using these flags, right?

bolshakov (approved):

PR #418, opened 2025-08-24
Prevent Stoplight failures when Redis is not reachable

### LINKED ISSUE #413: `Stoplight::DataStore::Failsafe#record_failure` fallback return value causes `NoMethodError` in `Stoplight::Light::GreenRunStrategy#record_error`

bolshakov:
Fixes #413

Example scenario:

```ruby
redis = Redis.new(url: "redis://561922f7-6b30-49d3-8148-324922d590d2:6379/0")
light = Stoplight('bug-report-2025-08-22', data_store: Stoplight::DataStore::Redis.new(redis))
light.run { raise 'foo' }
```

It raises "foo" error. Before the fix, it exhibited unexpected behavior.

💡Fixing this, I realized we could behave better when the data store is failing. Currently, we protect the underlying Redis data store with an in-memory-based circuit breaker. While it works for protecting the Redis data store itself, when Redis itself fails, the protected service becomes unprotected. **We should think of how to switch to in-Memory data store for protecting the service in such situations**

### DIFF
--- a/lib/stoplight/data_store/fail_safe.rb
+++ b/lib/stoplight/data_store/fail_safe.rb
@@ -48,13 +48,13 @@ module Stoplight
       end
 
       def get_metadata(config)
-        with_fallback(Metadata.new, config) do
+        with_fallback(EmptyMetadata, config) do
           data_store.get_metadata(config)
         end
       end
 
       def record_failure(config, failure)
-        with_fallback(nil, config) do
+        with_fallback(EmptyMetadata, config) do
           data_store.record_failure(config, failure)
         end
       end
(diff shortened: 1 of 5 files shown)


Lokideos (approved):
🚀

PR #459, opened 2025-10-07
Attach current config to a RedLight exception

Envek:
In some cases I would like to get specific light settings from a `Stoplight::Error::RedLight` exception that can be rescued at some top level where it can be unclear which circuit breaker was triggered.

E.g. to calculate background job retry time from specific light cool off time:

```ruby
class JobWithCustomRetry
  include Sidekiq::Job

  sidekiq_retry_in do |count, exception, _jobhash|
    case exception
    when Stoplight::Error::RedLight
      exception.config.cool_off_time * (count + 1)
    end
  end

  def perform(...)
  end
end
```

See [Sidekiq Error Handling](https://github.com/sidekiq/sidekiq/wiki/Error-Handling) for details on this example.

Right now only light name is available as exception message and, as far as I understand the code, it is not possible to get an existing instance with previously defined config from just a light name.

### DIFF
--- a/lib/stoplight/error.rb
+++ b/lib/stoplight/error.rb
@@ -5,6 +5,14 @@ module Stoplight
     Base = Class.new(StandardError)
     ConfigurationError = Class.new(Base)
     IncorrectColor = Class.new(Base)
-    RedLight = Class.new(Base)
+
+    class RedLight < Base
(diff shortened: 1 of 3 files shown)


bolshakov:
Addressed here #484

PR #460, opened 2025-10-07
Use aggregated buckets for memory data store - 82x speed up

bolshakov:
Instead of mimicking Redis data store storage schema, I changed the memory data store to pre-aggregate metrics into 1-second buckets. 

Benchmark: 

```ruby
# frozen_string_literal: true

require "benchmark/ips"
require_relative "../lib/stoplight"
require "redis"

Benchmark.ips do |b|
  b.report("After") { Stoplight("test", window_size: 5).run { 2 + 2} } # executed on this branch
  b.hold!("hold")
  b.report("Before") { Stoplight("test", window_size: 5).run { 2 + 2} } # executed on develop branch

  b.compare!
end
```

Benchmark results:

```
✗ ruby bench/compare.rb
ruby 3.2.4 (2024-04-23 revision af471c0e01) [arm64-darwin23]
Warming up --------------------------------------
               After     8.917k i/100ms
Calculating -------------------------------------
               After     82.353k (± 6.2%) i/s   (12.14 μs/i) -    410.182k in   5.007388s

Pausing here -- run Ruby again to measure the next benchmark...
➜  stoplight git:(feature/memory-buckets) ✗ gco develop
Switched to branch 'develop'
➜  stoplight git:(develop) ✗ ruby bench/compare.rb
ruby 3.2.4 (2024-04-23 revision af471c0e01) [arm64-darwin23]
Warming up --------------------------------------
              Before   238.000 i/100ms
Calculating -------------------------------------
              Before    999.821 (±14.5%) i/s    (1.00 ms/i) -      4.998k in   5.099942s

Comparison:
               After:    82352.9 i/s
              Before:      999.8 i/s - 82.37x  slower
```


bolshakov:
While this gives significant improvement, I believe we can do better. A few ideas to explore:
- tumbling window instead of running window -- very effective but sucks, especially for longer window sizes  
- hierachical buckets with different resolutions (1s, 5s, 30s, etc.)
- prefix sum storing cumulative sum for each second, should give us 0(1) for both reads and writes. Someting like this:
```ruby
# update 
@errors[timestamp] = @errors[timestamp - 1] + count_for_that_second

# reads 
sum_in_window = @errors[end_ts] - @errors[start_ts - 1]
```
- Instead of Hash, use a circular buffer. This could yield better performance in practice - iterating over an array may be more effective than iterating over a hash.

PR #469, opened 2025-10-11
feat: Introduce public config

Lokideos:
This PR aims to introduce public config, which will be exposed when users need to access additional information from our Lights.

We need it since we want to avoid direct access to our [Stoplight::Light::Config](https://github.com/bolshakov/stoplight/blob/8c7208dd5f060bab7840c1769370376ede333e82/lib/stoplight/light/config.rb), since it will prevent us from easily changing Stoplight internals.

For now, we expose only the Time when the Light turns Yellow. We will provide more info in the public config in the future.


bolshakov:
Addressed here #484

PR #471, opened 2025-10-12
Timestamp metrics at recording time, not event time

bolshakov:
Standardize timestamp handling by using the current wall-clock time at the moment of recording, rather than accepting timestamps from callers.

### Before

The code used `failure.time` (when the error occurred) or accepted `request_time` from callers, which could lead to inconsistencies if there's latency between the failure occurrence and recording, and therefore confusion about which timestamp to use.

### After

All events are timestamped with the actual time they were recorded to the data store to ensure windows are calculated based on when the data store saw the event. This:
- Prevents back-dating/future-dating of events -- the events are recorded in order of arrival
- Makes testing cleaner (use Timecop to control time globally)

---

**Context**: This is a prerequisite for implementing more efficient memory data store operations. The current implementation filters all events within a window (_O(n)_), but with guaranteed monotonic insertion, we can maintain running sums that make window queries at _O(log(n))_ time.

### DIFF
--- a/lib/stoplight/data_store/memory.rb
+++ b/lib/stoplight/data_store/memory.rb
@@ -30,10 +30,10 @@ module Stoplight
       # @return [Stoplight::Metadata]
       def get_metadata(config)
         light_name = config.name
+        current_time = self.current_time
+        recovery_window = (current_time - config.cool_off_time)..current_time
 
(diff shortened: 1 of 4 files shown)


Lokideos (approved):
🚀

PR #481, opened 2025-10-20
feat: Add nonce to script tags in stoplight admin

Tony-Chen-Persona:
Adding an option to set a nonce to the script tags in the stoplight admin gem for content security policy reasons.

The nonce can be set by extending the `Stoplight::Admin` class.


bolshakov:
@Tony-Chen-Persona, thank you for your contribution. Please make sure all CI steps pass so we can proceed with the merge

bolshakov (approved):

PR #495, opened 2025-10-30
Extract request tracking logic into dedicated tracker classes

bolshakov:
Introduces BaseTracker, RequestTracker, and RecoveryProbeTracker to separate concerns between execution strategies and state management.

- Strategies focus on execution, trackers handle state
- Recording logic can be tested independently (this simplifies tests dramatically)

### DIFF
--- a/lib/stoplight/domain/strategies/yellow_run_strategy.rb
+++ b/lib/stoplight/domain/strategies/yellow_run_strategy.rb
@@ -46,7 +46,7 @@ module Stoplight
         # @return [Object] The result of the code block if successful.
         # @raise [Exception] Re-raises the error if it is not tracked or no fallback is provided.
         def execute(fallback, metadata:, &code)
-          enter_recovery_if_needed(metadata)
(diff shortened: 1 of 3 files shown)


lib/stoplight/domain/strategies/red_run_strategy.rb
```diff
@@ -10,6 +10,14 @@ module Strategies
       #
       # @api private
       class RedRunStrategy < RunStrategy
+        # @!attribute [r] config
+        #   @return [Stoplight::Domain::Config] The configuration for the light.
+        protected attr_reader :config
+
+        def initialize(config:)
+          @config = config
```
Lokideos: Do we want to add super here and in other strategies to make strategies more extensible in the future?

bolshakov: Actually, I removed the `RunStrategy#initialize` method because different strategies have different dependencies. We could reconsider his if a common initialization logic arises

PR #551, opened 2025-12-16
Move State to Dedicated Storage Classes

bolshakov:
Continues storage abstraction separation by extracting state management from DataStore into dedicated per-circuit storage instances.

Previously, we introduced the `Domain::Storage::State` interface. This PR implements the actual storage classes following the same pattern as recovery locks.

**Before:** DataStore handled state for ALL circuits:

```ruby
data_store.set_state(config, state)
snapshot = data_store.get_state_snapshot(config)
first_to_transition = data_store.transition_to_color(config, Color::RED)
```

**After:** each circuit has its own state storage instance:

```ruby
state_store = Memory::State.new(clock:, cool_off_time:)  # Per circuit
state_store.set_state(state)  # No config needed - knows its circuit
snapshot = state_store.state_snapshot
first_to_transition = state_store.transition_to_color(Color::RED)
```

Also introduces `Domain::Clock` interface with `Infrastructure::SystemClock` implementation, making time dependencies explicit and testable.

The Redis implementation is composed with `FailSafe::State` decorator for automatic memory failover, wired in `LightBuilder`.

### DIFF
--- a/lib/stoplight/infrastructure/storage/fail_safe/state.rb
+++ b/lib/stoplight/infrastructure/storage/fail_safe/state.rb
@@ -26,9 +26,9 @@ module Stoplight
           #   @return [Stoplight::Light] The circuit breaker used to handle store failures.
           private attr_reader :circuit_breaker
 
-          # @param primary_store [Stoplight::Domain::Storage::RecoveryLock]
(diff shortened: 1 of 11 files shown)


Lokideos (approved):
🚀

PR #571, opened 2025-12-30
Support Redis Cluster

fonsecas72:
Trying to add support for redis cluster

### DIFF
--- a/lib/stoplight/infrastructure/data_store/redis.rb
+++ b/lib/stoplight/infrastructure/data_store/redis.rb
@@ -24,10 +24,15 @@ module Stoplight
           # Generates a Redis key by joining the prefix with the provided pieces.
           #
           # @param pieces [Array<String, Integer>] Parts of the key to be joined.
+          # @param cluster_mode [Boolean] Whether to use hash tags for cluster compatibility
           # @return [String] The generated Redis key.
           # @api private
-          def key(*pieces)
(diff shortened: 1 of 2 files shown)


bolshakov:
Hey Hugo! This is an excellent contribution - the implementation is solid, and I appreciate you tackling the Redis Cluster compatibility problem.

That said, I've decided not to merge this, not because there's anything fundamentally wrong with it, but because of architectural changes planned for Stoplight 6.0.

You've probably noticed the tension with supporting the `#names` method. That's because Stoplight currently discovers circuit breaker names through Redis SCAN, which is inherently problematic in cluster mode since keys are distributed across nodes. Your `scan_all_cluster_nodes` workaround is clever, but it's fighting against the underlying architecture.

In 6.0, we're replacing this approach with an explicit registry, so querying light names will be a matter of issuing a single Redis command regardless of cluster mode. This also means supporting hash-tagged keys becomes trivial, in fact, I think we should use them by default ([see these  changes](https://github.com/bolshakov/stoplight/pull/584)):

```ruby
def key(*pieces) = [:stoplight, :v6, system_id, "{#{light_id}}", *pieces].join(":")
```

Thanks again for the contribution! If you're interested in helping with the 6.0 work, I'd welcome your involvement.

fonsecas72:
@bolshakov 
image

PR #601, opened 2026-02-03
Fix incorrect failure count in Admin Dashboard

### LINKED ISSUE #599: Admin dashboard shows incorrect failure count
**Describe the bug**
The admin dashboard shows the wrong number of failures for a circuit. After triggering 3 consecutive failures (enough to trip the circuit), the dashboard still shows **"Failures: 1"** instead of **"Failures: 3"**.

**To Reproduce**
1. Configure Stoplight with Redis as the data store and ensure Redis is running.
2. Define a circuit with threshold 3, e.g. `Stoplight('my-circuit', threshold: 3)`.
3. Trigger 3 consecutive failures (e.g. run the circuit 3 times with a block that raises).
4. Mount the Stoplight Admin and open the dashboard (e.g. `/stoplights`).
5. Find the circuit card and check the "Failures:" value.

**Expected behavior**
The dashboard should show **"Failures: 3"** (or whatever the actual consecutive failure count is) so it matches the number of failures that led to the circuit opening.

**Screenshots**
Image

**Environment:**
 - OS: Linux (Ubuntu 24.10)
 - Ruby: 3.4.2  
 - Stoplight version: 5.7.0
 - Data store: Redis
 - Redis: 7.4.1 (running via Docker with redis:alpine image)

**Additional context**
- Reproduction requires **Redis** as the data store. With the default in-memory store, the admin does not run (it raises that a persistent store is required), so the bug is only visible with Redis.
- The circuit itself behaves correctly (it opens after 3 failures). The issue is only the number displayed on the dashboard.

oLucasAguilar:
Fixes #599 

### Motivation

The Admin Dashboard was displaying "Failures: 1" regardless of the actual number of consecutive errors that tripped the circuit.

### Solution

Previously, the view calculated the count based on the `failures` array.
However, this array contained only the most recent error (at most 1 element).

This was incorrect because it didn't reflect the actual consecutive failure count that the circuit breaker uses for its logic. I have updated the class to fetch the actual `consecutive_errors` metric and pass it through to the view.

### Reproduction

I tested the solution using the reproduction script provided in the issue (see #599 for the full script).
This script triggers 3 failures to trip a circuit with a threshold of 3.

**Before:**
image

**After:**
(count fixed)
 image

### DIFF
--- a/lib/stoplight/admin/lights_repository.rb
+++ b/lib/stoplight/admin/lights_repository.rb
@@ -75,7 +75,8 @@ module Stoplight
           name: name,
           color: state_snapshot.color,
           state: state_snapshot.locked_state,
-          failures: [metrics.last_error].compact
+          failures: [metrics.last_error].compact,
+          failure_count: metrics.consecutive_errors
         )
       end
 
--- a/lib/stoplight/admin/lights_repository/light.rb
+++ b/lib/stoplight/admin/lights_repository/light.rb
@@ -37,16 +37,23 @@ module Stoplight
         # @dynamic failures
         attr_reader :failures
 
+        # @!attribute failure_count
+        #   @return [Integer]
+        # @dynamic failure_count
+        attr_reader :failure_count
+
         # @param name [String]
         # @param color [String]
         # @param state [String]
         # @param failures [<Stoplight::Failure>]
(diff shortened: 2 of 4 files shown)


bolshakov (approved):
Great contribution, thank you!

PR #668, opened 2026-05-01
refactor: Slightly adjust Admin panel design

Lokideos:
This PR attempts to slightly improve Admin panel design:

1. Cards alignment on the grid. Currently cards behave a little bit weird if we have only two of them. I think that if we stick them to the left side it would look better.

| Before | After |
| ------ | ----- |
| Screenshot 2026-05-01 at 17 27 14 | Screenshot 2026-05-01 at 17 28 06 |

2. Our background cuts after one screen. So if user have a lot of lights and scrolls down background changes to plain white background.


bolshakov (approved):
Great way to celebrate labor day 🚀

PR #681, opened 2026-05-23
Optimize light configuration and caching

bolshakov:
Improved `System#light` caching performance using pre-computed configuration digests and added better error messages with source locations.

-  Pre-compute digest during  `LightConfigurationDsl` initialization
-  Implement `#hash` for traffic/recovery control strategies -- allows comparing configurations effectively
- Cache UNDEFINED constant (saves `Undefined.instance` lookup overhead) - there is more room for improvement here, but it does not play well with RBS.
- Improve configuration conflict errors. Previously error message included both configurations, not it includes both call sites.
- Split DSL classes by use case
   - `LightConfigurationDsl`: Optimized for light caching with pre-computed digest
   - `SystemConfigurationDsl`: For system-level configuration
   - `LegacyConfigurationDsl`: For backward compatibility, used outside of system, will be removed eventually.

### Performance impact

More effective lights cashing enabled a massive performance gain on light lookup.

```
ruby 3.2.4 (2024-04-23 revision af471c0e01) [arm64-darwin23]
Warming up --------------------------------------
               after    70.405k i/100ms
Calculating -------------------------------------
               after    690.021k (± 3.8%) i/s    (1.45 μs/i) -      3.450M in   5.007544s

Comparison:
               after:   690021.2 i/s
              before:   251477.5 i/s - 2.74x  slower
```

When we switch to "system lights", a hottest path will have a very little overhead for light instantiation:

```ruby
Stoplight(name).run {. ... } # <- Faster lookup
```

### DIFF
--- a/lib/stoplight/wiring/system.rb
+++ b/lib/stoplight/wiring/system.rb
@@ -83,7 +83,7 @@ module Stoplight
         traffic_control: T.undefined,
         traffic_recovery: T.undefined
       )
-        light_config = ConfigurationDsl.new(
+        light_config = LightConfigurationDsl.new(
           name:,
           cool_off_time:,
           threshold:,
--- a/lib/stoplight/wiring/system.rb
+++ b/lib/stoplight/wiring/system.rb
@@ -83,7 +83,7 @@ module Stoplight
         traffic_control: T.undefined,
         traffic_recovery: T.undefined
       )
-        light_config = LightConfigurationDsl.new(
+        light_dsl = LightConfigurationDsl.new(
           name:,
           cool_off_time:,
           threshold:,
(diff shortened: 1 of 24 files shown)


lib/stoplight/wiring/system.rb
```diff
@@ -98,26 +93,37 @@ def light(
...
+            source_line = caller(6, 1)&.first
+            [
+              LightFactory.new(
+                system: self,
+                config: light_dsl.configure!(system_config)
+              ).build,
+              config_digest,
+              source_line
+            ]
           end
```
Lokideos, line 124: Since we only interested in light, maybe we want to return only light here? Something like:
```ruby
def light(
  name,
  cool_off_time: T.undefined,
  threshold: T.undefined,
  recovery_threshold: T.undefined,
  window_size: T.undefined,
  tracked_errors: T.undefined,
  skipped_errors: T.undefined,
  traffic_control: T.undefined,
  traffic_recovery: T.undefined
)
  light_dsl = LightConfigurationDsl.new(
    name:,
    cool_off_time:,
    threshold:,
    recovery_threshold:,
    window_size:,
    tracked_errors:,
    skipped_errors:,
    traffic_control:,
    traffic_recovery:
  )
  config_digest = light_dsl.digest
  light = lights.compute(name) do |existing|
    if existing
      existing_light, existing_digest, existing_source_line = existing

      if config_digest != existing_digest
        raise Stoplight::Error::ConfigurationError, <<~MSG
                Light `#{name}` already registered with different configuration.

                Original registration: #{existing_source_line}
                Current attempt: #{caller(6, 1)&.first}

                Lights must have consistent configuration across all call sites.
              MSG
      end

      existing_light
    else
      LightFactory.new(
        system: self,
        config: light_dsl.configure!(system_config)
      ).build
    end
  end

  light
end
```

Lokideos (approved):
Stellar work as always 🚀

lib/stoplight/wiring/system.rb
bolshakov: This deconstruction won't work if the block returns only lights

    existing_light, existing_digest, existing_source_line = existing


Block result is atore in the hash because we need it to compare current digest with existing digest. If we don't return all theee elements we won't have the digest to compare with and won't have the line trace to show in the error

PR #685, opened 2026-05-25
chore: Set favicon for Stoplight Admin Panel

Lokideos:
This PR aims to set default favicon for Stoplight Admin Panel.

If we mount Stoplight Admin panel to Rails app, we get annoying `ActionController::RoutingError (No route matches [GET] "/favicon.ico")` error. While it does not prevent us from using Admin Panel, we could fix it by adding a simple favicon to Stoplight Admin Panel assets.


bolshakov (approved):
Nice touch!

PR #688, opened 2026-05-26
Remove #with and #with_* methods

bolshakov:
This PR removes `Light#with` and a bunch of `Light#with_*` methods -- all were deprecated. This introduces a breaking change into a develop branch marking the start of preparation for the major release. 

I temporarily mark `Light#==` as pending. It will be fixed on the following changes. After switching to system light by default, light will be compared by identity. So I don't want to implement a termorary solution just to remove it on the next step.

Next steps:
- Unify two  LightFactory and LightBuilder hierarchies which are now redundant


Lokideos (approved):
🚀

PR #689, opened 2026-05-26
Unify Factory with Builder

bolshakov:
Simplifies the wiring by merging `LightBuilder` into `LightFactory`.

The Factory/Builder split was overcomplicated:
- `LightFactory` only parsed configuration and delegated to builder
- `LightBuilder` did all actual construction work
- 1:1 mapping between factory and builder

This PR removes `LightBuilder` moving `LightFactory#with` to `LightBuilder#with`.


sig/_private/stoplight/domain/ports/light_factory.rbs
Lokideos: Potentially we might want to get rid of this interface since we use it only here.

Lokideos (approved):
🚀

PR #696, opened 2026-06-05
Separate light and system settings

### LINKED ISSUE #482: Remove error_notifier from public API
The `error_notifier` option has never been documented in the README and may cause confusion for users.

In the Stoplight function documentation, it’s currently described as:

```
# @option settings [Proc] :error_notifier A proc to handle error notifications.
```

This phrasing could misleadingly suggest that the proc reports any errors occurring during the execution of user code. However, this is not the case. The `error_notifier` is used internally by Stoplight to report internal failures - for example, when the Redis data store fails. In such cases, Stoplight falls back to the in-memory data store while using this notifier to log or report the internal error.

To make its purpose clearer and prevent misuse, I propose moving this configuration option to the global configuration only, disallowing per-light configuration.

```ruby
  config.error_notifier = ->(error) { warn error }
end
```

This makes it explicit that the `error_notifier` applies globally and is intended solely for internal error reporting.

bolshakov:
Closes #482

Stoplight has two fundamentally different kinds of configuration that were previously mixed into the same parameter list:

**Infrastructure configuration** -  how circuit breaker state is stored and reported:

- `data_store` - where failure counts and state live (Memory, Redis)
- `notifiers` — who gets told when a light changes color
- `error_notifier` — who gets told when Stoplight itself has an internal error

**Behavior configuration** — *when* a circuit breaks and recovers:

- `threshold`, `cool_off_time`, `window_size`, `recovery_threshold`
- `tracked_errors`, `skipped_errors`
- `traffic_control`, `traffic_recovery`

These have different concerns. Infrastructure is a singleton operational decision - the whole application shares one Redis connection, one set of Slack notifiers. Behavior is per-circuit — the payment gateway trips at 5 errors in 60 seconds, the recommendation engine can tolerate more. Previously, Stoplight let you mix them freely, which caused a real class of bugs.

The footgun this eliminates:

```ruby
Stoplight("Payment Service", data_store: Stoplight::DataStore::Memory.new).run { charge_card }
```

Every call to `Stoplight()` here creates a fresh `Memory` store. Since `Memory` stores state in a plain hash, each invocation starts with zero failures recorded. The circuit breaker never trips — it silently discards all failure history on every call. The behavior just doesn’t work.

This PR makes it a hard error. `data_store:`, `notifiers:`, and `error_notifier:` are no longer valid parameters to `Stoplight()`. You configure infrastructure once at boot:

```ruby
# initializer
  config.data_store = Stoplight::DataStore::Redis.new(redis)
  config.notifiers  = [Stoplight::Notifier::Logger.new(logger)]
end

# Anywhere in your app - behavior only
Stoplight("Payment Service", threshold: 5, window_size: 60).run { charge_card }
```

The right configuration lives in the right place, and the wrong thing is impossible to express.

## Migration

The change is mechanical for the vast majority of users:

```ruby
# Before
light = Stoplight("name", data_store: my_store, notifiers: [my_notifier], threshold: 5)

# After
  config.data_store = my_store
  config.notifiers  = [my_notifier]
end
light = Stoplight("name", threshold: 5)
```

## Escape Hatch: Systems

If you have genuinely separate infrastructure boundaries - say, one Redis for payment circuits and another for background jobs — `Stoplight.system` lets you own that explicitly:

```ruby
payments_system = Stoplight.system("payments", data_store: Stoplight::DataStore::Redis.new(payments_redis))
payments_system.light("charge", threshold: 3).run { charge_card }
```

Most applications will never need this.


bolshakov:
We discussed this change and quite confident about `data_store` and `error_notifier` removal. However, we have doubts regarding the `notifiers` change as some users may have built business logic around it. This never was an intended use, but yet we don't want to break such cases. 

We decided to merge it now, but review this decision before the release. We encourage users to speak up if this affects them.

PR #711, opened 2026-06-29
Add Registry storage abstraction

bolshakov:
## Summary

Until now, Stoplight had no concept of a registry - a place to record which named lights
belong to a system. The upcoming change that makes all lights go through a `System` needs
to enumerate lights by system (for the admin panel), but the existing `DataStore` interface 
has no such concept.

This PR introduces `Domain::_Registry` as a dedicated interface for that single
responsibility, along with three concrete implementations that mirror the pattern already
used for other storage types:

- `Infrastructure::Memory::Storage::Registry` - a no-op stub (in-memory lights do not
  persist registrations across restarts, so tracking them is unnecessary)
- `Infrastructure::Redis::Storage::Registry` - stores light names in a Redis hash,
  keyed by `SystemKeySpace`
- `Infrastructure::FailSafe::Storage::Registry` - wraps a primary registry with
  circuit-breaker fallback to the memory stub
 
No existing code is wired to these classes yet. The wiring follows in the next PR.

### DIFF
--- a/lib/stoplight/infrastructure/fail_safe/storage/registry.rb
+++ b/lib/stoplight/infrastructure/fail_safe/storage/registry.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+module Stoplight
+  module Infrastructure
+    module FailSafe
+      module Storage
+        class Registry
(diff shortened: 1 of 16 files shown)


Lokideos (approved):
🚀

PR #712, opened 2026-06-29
Systems (Part 1) - wire System with injected failover and registry

bolshakov:
Before this change, `Wiring::System` constructed its own failover system lazily on first use, and `Stoplight.light` / `Stoplight.system` managed a loose collection of global state (a `@systems` map, `@default_config`, and `@default_light_factory` as separate ivars). There was no place that owned the wiring of a system's external dependencies.

- Introduce `Wiring::GlobalState` as the single place that constructs and owns every `System`. It builds the failover system and the registry before passing them into `System.new`, so each system's dependencies are explicit at construction time rather than materialised ad-hoc.
- `System#initialize` now takes `failover_system:` and `registry:` as required keyword args. The lazy `failover_system` private method is gone; `GlobalState` is the factory.
- `GlobalState#create_registry` wires up `Infrastructure::FailSafe::Storage::Registry` (backed by Redis + Memory failover) for Redis-backed systems, and a plain `Memory::Storage::Registry` otherwise. The failover system's `"registry"` light is used as the circuit breaker, so registry failures are protected by the same failover path as other Redis operations.
- `Stoplight.configure` now creates a single `GlobalState` instance instead of three separate ivars, simplifying reset and configured-state tracking.
- `System::LightFactory` drops `system:` in favour of `system_name: String` — it only needed the name for key-space construction, not the full object.
- Add `System::Storage` — a thin adapter that maps `(config) -> state_snapshot / metrics_snapshot / delete` by delegating to the appropriate `LightFactory`. Used by the admin interface to read cross-system state.


lib/stoplight/wiring/global_state.rb
```diff
@@ -0,0 +1,70 @@
...
+      end
+
+      private
+
+      def create_registry(config, failover_system)
+        case config.data_store
+        when DataStore::Redis
+          Infrastructure::FailSafe::Storage::Registry.new(
+            primary_registry: Infrastructure::Redis::Storage::Registry.new(
+              redis: redis(config),
```
Lokideos:
```suggestion
              redis: config.data_store.redis,
```

bolshakov: Good catch, fixed

PR #713, opened 2026-06-30
We all go agentic! 

bolshakov:
Well, regardless agentic sentiment, the repository got used by agents and it's better to help them navigate rather not. This PR adds:

- `AGENTS.md` with general library information and opinionated nudge agents toward TDD, useful commands 
- `CLAUDE.md` just references `AGENTS.md` 
- `docs/architecture.md`: rewritten, more terse content, less fluff, less concrete code examples that tend age fast.
- `docs/testing.md`: detailed expectation regarding testing - what and how should be covered. More specific TDD instructions
- `docs/redis_and_storage.md` - how we work with redis, fail safe storage, mentions migration out of `_DataStore` interface.
- , `docs/types_and_rbs.md` - describes how to work with RBS, where to put files, etc. 
- `.claude/rules/` - claude-specific symlinkes to the above files, allows claude load them on demand 
- `.claude/settings.json`:  wired standardrb format (`PostToolUse`), lint and steep gates (Stop), 

At the end of the day, I hope this should allow agent to contribute better code and humans understand our codebase and processes better.


bolshakov:
@Lokideos I decided it would be easier to just vendor the TDD skill from https://github.com/obra/superpowers

This is a more concise workflow then my initial rudimentary attempt to steer toward TDD without describing it.

AGENTS.md
Lokideos, line 1: Let's add guidance for mandatory comments for Stoplight public API (which should include examples of usage).

docs/data_storages.md
Lokideos, line 1: Maybe we should rename it to `data_storages` since we plan to implement more data stores.

Lokideos (approved):
Stellar work! 🚀 
I think we might also want to think about gem discoverability from LLMs ^_^

PR #721, opened 2026-07-04
Stoplight Telemetry

bolshakov:
## Summary

Add a small, typed, in-process event bus to Stoplight that emits value-object events at the key moments of a light's lifecycle: every run, every state transition, every recovery probe, every lock change, and light registration. 

The core only emits. Aggregation, buffering, and export are subscriber concerns out of the scope of this implementation.

## Motivation

Today Stoplight records failures and notifies **only on color change**. Everything else is invisible:

- Successful runs and their latency are discarded. Degradation detection is impossible because no duration is captured anywhere.
- The only extension point, `Notifier`, sees just `(from_color, to_color, error)` - too narrow for observability dashboards, alerting, or metrics pipelines.

Stoplight already wraps every external dependency call. A typed event stream turns it into a **dependency-reliability sensor** without asking users to instrument anything.

## What this opens up?

None of these ship in this proposal. The point is that the event stream adds an extension point that is sufficient for all of them, built by us or by anyone else as a subscriber gem:

- **Metrics pipelines.** Traffic rate, error rate, and latency percentiles are all  reconstructible from the `RunCompleted` stream alone (count, error count, blocked count,  duration histogram). An OpenTelemetry or StatsD exporter is a subscriber, nothing more.
- **Incident timelines.** The `StateTransitioned` stream is, by contract, the complete color timeline: breaches, recovery attempts, and manual locks, including the lock someone set during an incident and forgot.
- **Fleet inventory.** `LightRegistered` reports every light each process knows about, together with its full serializable configuration — which lights exist, with which  policies, in which processes, before anything fails.
- **Recovery observability.** `RecoveryProbeCompleted` shows a dependency coming back (probe latency, progress toward the recovery threshold) instead of a silent red -> green jump.
- **The notifier system becomes a subscriber.** The existing notifiers keep working,  reimplemented as a bridge that subscribes to `StateTransitioned` - one delivery mechanism instead of two parallel ones.

## Design principles

1. **Core emits, subscribers decide.** No buffering, aggregation, or I/O in the domain layer. Delivery is synchronous on the calling thread. Anyone who wants async builds it into their subscriber (queue-and-return), not into the bus.
2. **Zero cost when nobody listens — per event class.** Emission of an event class with no subscribers allocates nothing: no payload, no envelope, no block object. .
4. **Make impossible states unrepresentable.** Every field permutation an event can express  is a state the domain can reach. This is why transitions are five classes instead of one class with a `reason` enum and a handful of nilable fields.
5. **Events are process-local observations.** They describe what *this process* observed or did, not the globally shared circuit state. 
6. **Typed end to end.** Subscription overloads narrow the payload to the subscribed class,  emission overloads pin the payload type to the emitted class. 

### Envelope

Every delivered event is wrapped in an `Envelope` carrying the coordinates the emission site
knows:

```rbs
class Envelope[out P < event] < Data
  attr_reader system_name: String   # lights live in named systems; "default" for most users
  attr_reader light_name: String
  attr_reader occurred_at: Time
  attr_reader payload: P
end
```

The envelope is generic over its payload, so a filtered subscription receives `Envelope[RunCompleted]`, not `Envelope[Event]`.

### Events

Eight event classes. All are frozen `Data` value objects under `Stoplight::Domain::Telemetry`.

| Event | Volume | Fields |
|---|---|---|
| `RunCompleted` | every `Light#run` | `outcome` (`:success/:failure/:blocked`), `color`, `duration_ms` (nil only when blocked), `failure?`, `fallback_used`, `retry_after?` |
| `TrafficBreached` | rare | `from_color: green`, `to_color: red`, `policy`, `failure`, `metrics` |
| `RecoveryStarted` | rare | `from_color: red`, `to_color: yellow`, `breached_at` |
| `RecoverySucceeded` | rare | `from_color: yellow`, `to_color: green`, `policy`, `metrics` |
| `RecoveryFailed` | rare | `from_color: yellow`, `to_color: red`, `policy`, `failure?`, `metrics` |
| `LockChanged` | rare | `from_color`, `to_color`, `from_state`, `to_state`, `source` |
| `RecoveryProbeCompleted` | low | `outcome`, `duration_ms`, `failure?`, `progress` (recovery metrics after this probe) |
| `LightRegistered` | once per (process, system, light) | `settings` (full serializable config) |

### Transitions

The five transition events form a closed union over the state machine's edges:

```rbs
type state_transitioned = TrafficBreached | RecoveryStarted | RecoverySucceeded | RecoveryFailed | LockChanged
```

Closed on purpose: new traffic/recovery *policies* never add union variants -- they self-identify through the open `policy` string inside the matching variant (`"error_rate"`, `"consecutive_errors"`, a user's custom policy). 

RBS unions have no runtime constant, so the union gets a runtime name via a marker module: `StateTransitioned` is included by every variant and carries no behavior. It exists so a subscription filter can dispatch on the whole union.

## Usage

The bus is a `System` component, wired like storage and notifiers are today: each system has its own bus, so tests and multi-system setups are isolated by construction. 

```ruby
# 1. Firehose: every event, including types added in future versions.
telemetry.subscribe do |envelope|
  case (payload = envelope.payload)
  when Telemetry::RunCompleted
    # payload narrowed to RunCompleted here (Steep understands the case)
  when Telemetry::StateTransitioned
    # any of the five transition variants
  end
end

# 2. One event class: the payload arrives pre-narrowed.
telemetry.subscribe(Telemetry::RunCompleted) do |envelope|
  run = envelope.payload # : Envelope[RunCompleted], no casting
  STATS.increment("stoplight.runs", tags: ["light:#{envelope.light_name}", "outcome:#{run.outcome}"])
  STATS.timing("stoplight.duration_ms", run.duration_ms) if run.duration_ms
end

# 3. The transition union via its marker module: the complete color timeline.
subscription = telemetry.subscribe(Telemetry::StateTransitioned) do |envelope|
  Rails.logger.warn("[stoplight] #{envelope.light_name}: " \
    "#{envelope.payload.from_color} -> #{envelope.payload.to_color}")
end

telemetry.unsubscribe(subscription)
```

## Emitting: the producer side

Producers (run strategies, trackers, `Light`) don't see the bus directly. They get an `Emitter` - one per light, bound to its coordinates at construction time in wiring, injected the same way notifiers are today:

```ruby
emitter.emit(Telemetry::RunCompleted) do
  Telemetry::RunCompleted.new(
    outcome: :success, color: :green, duration_ms: elapsed_ms,
    failure: nil, fallback_used: false, retry_after: nil
  )
end
```

The shape is intentional : the class argument runs the guard that enables us to yield the block only if someone is interested in this type of event. If nothing is subscribed to that class, the payload is never constructed. Per-class overloads pin the block's return type to the class argument  so  `emit(RunCompleted) { TrafficBreached.new(...) }` is a Steep error.

## Current notifiers become a bridge

The existing notifier API is reimplemented as the first subscriber: a bridge that listens to `StateTransitioned` and translates to today's `notify(config, from_color, to_color, error)` calls. User-visible notifier behavior doesn't change; the gem does not have two parallel delivery mechanisms. The bridge is also the reference subscriber.

## Other considerations 

- The public accessor for the bus (`Stoplight.telemetry` delegating to the default system not expressed in the PR but implied..
- `Settings#tracked_errors`/`skipped_errors` matchers by constant name, which requires matchers to be named constants.  Not it allows broader interface (e.g. a `Proc`) this will be deprecated, only constants accepted.

## Follow-up PRs

The whole thing is implemented in three PRs that could be reviewed sequentially:

1. https://github.com/bolshakov/stoplight/pull/723
2. https://github.com/bolshakov/stoplight/pull/724
3. https://github.com/bolshakov/stoplight/pull/726

### DIFF
--- a/sig/_private/stoplight/domain/ports/telemetry_publisher.rbs
+++ b/sig/_private/stoplight/domain/ports/telemetry_publisher.rbs
@@ -1,0 +1,14 @@
+module Stoplight
+  module Domain
+    # Producer-facing side of the telemetry bus, injected into Telemetry::Emitter. End users never
+    # see it: they get the consumer side (_Telemetry) from +_System#telemetry+.
+    interface _TelemetryPublisher
+      # True when any subscriber (firehose, exact class, or via the StateTransitioned marker) would
+      # receive this event class. The Emitter guards on it before building anything.
+      def subscribed?: (Telemetry::event_class) -> bool
+
+      # Delivers to every interested subscriber, synchronously, in subscription order.
+      def publish: (Telemetry::Envelope[Telemetry::event]) -> void
+    end
+  end
+end
--- a/sig/_private/stoplight/domain/telemetry.rbs
+++ b/sig/_private/stoplight/domain/telemetry.rbs
@@ -1,0 +1,62 @@
+module Stoplight
+  module Domain
+    # In-process telemetry bus.
+    #
+    # Contract:
+    # - Events are process-local observations: they describe what this process observed or did, not the
+    #   globally shared circuit state.
+    # - Every color change this process observes or effects emits exactly one +state_transitioned+ event. That
+    #   stream alone reconstructs the color timeline (breaches, recovery, and locks included).
+    # - Events are traffic-driven - no traffic, no events.
+    # - Handlers run synchronously on the calling thread and are error-isolated:
+    #   a raising handler is routed to the error notifier and never affects the run.
+    # - Emission allocates nothing when nothing is subscribed to that event class - a transitions-only
+    #   subscriber keeps RunCompleted emission free.
+    # - Publish pays for interested handlers only: no filter matching happens at publish time
+    #   (interest is resolved at subscribe time), and subscriptions to other event classes add
+    #   nothing. The handlers themselves run inline, so their cost belongs to the subscriber.
+    #   Subscribing mid-publish takes effect on the next publish.
+    class Telemetry
+      # The consumer side (subscribe/unsubscribe), what +_System#telemetry+ returns.
+      include _Telemetry
+      # The producer side (publish/subscribed?), what wiring injects into Emitter.
+      include _TelemetryPublisher
+
+      # @param error_notifier receives any StandardError a subscriber's handler raises.
+      # @param max_subscriptions caps the number of live subscriptions - #subscribe raises
+      #   Stoplight::Error::TooManySubscriptions once reached.
+      def initialize: (error_notifier: error_notifier, ?max_subscriptions: Integer) -> void
+
+      # Run outcomes reflect the light's accounting, not the caller's experience: an untracked (skipped) error
+      # emits +:success+ with +failure.tracked = false+.
+      type run_outcome = :success | :failure | :blocked
+      type probe_outcome = :success | :failure
+
+      # Closed union over the state machine's edges. New traffic/recovery policies never add variants. They
+      # self-identify via the open +policy+ string inside the matching variant. Every variant includes
+      # the +StateTransitioned+ marker module - the union's runtime name.
+      type state_transitioned =
+        TrafficBreached |
+        RecoveryStarted |
+        RecoverySucceeded |
+        RecoveryFailed |
+        LockChanged
+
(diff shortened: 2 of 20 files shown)


Lokideos (approved):
🚀

PR #739, opened 2026-07-11
Add Stoplight.register / Stoplight.light for reusable, faster-looked-up lights

bolshakov:
Calling `Stoplight("name", ...)` at every call site works well for a handful of lights. As
an app grows, repeating the same settings everywhere makes them easy to drift out of sync,
and there's no single place listing what lights exist.

- Adds `Stoplight.register(name, ...)` to define a light's configuration once (e.g. in an
  initializer) and `Stoplight.light(name)` to look it up by name anywhere else, without
  repeating settings. `Stoplight.light` raises the new
  `Stoplight::Error::UnregisteredLightError` if the name was never registered.
- `Stoplight(name, ...)` still works exactly as before and now registers under the hood, so
  existing call sites are unaffected - registration is additive, not a replacement.
- `Stoplight.light` is also faster: `Stoplight.register`/`Stoplight(name, ...)` re-validates
  the configuration digest on every call, while `Stoplight.light` is a plain lookup. It's 
  ~9-10x more iterations/sec compared to `Stoplight()` calls..

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,7 +48,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        light-creation: [ 'Stoplight()', 'System#light' ]
+        light-creation: [ 'Stoplight()', 'System#register' ]
         data-store: [ 'Redis' ]
         data-store-image:
           - 'redis:7.4'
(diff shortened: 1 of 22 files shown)


Lokideos (approved):
Great improvement 🚀 
I'd rename `Configuration` section in README to `Advanced Configuration` though since this is another complexity point for Stoplight configuration and I'd rather keep it simple 🙏

bolshakov:
We already have Advance configuration section. Let me move registration there - make perfect sense

PR #749, opened 2026-07-12
fix: deduplicate recovery probe notifications

### LINKED ISSUE #748: `Tracker::RecoveryProbe#recover` discards `transition_to_color`'s dedup boolean and notifies unconditionally
`Domain::Tracker::RecoveryProbe#recover`:

```ruby
  state_store.transition_to_color(to_color)   # return value discarded
  metrics_store.clear
  info = LightInfo.new(name: config.name)
  notifiers.each do |notifier|
    notifier.notify(info, from_color, to_color, nil)
  end
```

Its sibling, `Domain::Tracker::Request#transition_to_red`, handles the equivalent green -> red trip correctly:

```ruby
  if state_store.transition_to_color(Color::RED)
    info = LightInfo.new(name: config.name)
    notifiers.each { |notifier| notifier.notify(info, Color::GREEN, Color::RED, exception) }
  end
```

`State#transition_to_color`'s own doc comment documents why that guard exists:

> Manages light state transitions and ensures notification deduplication through atomic transition detection. .. enabling exactly-once notification semantics.
>
> Multiple threads may call this concurrently
> if state.transition_to_color(Color::RED)
>   Only one thread reaches here - send notification
>   notifier.notify(circuit_name, :opened)
> end

`#recover` doesn't honor that contract. In the normal path the recovery lock serializes probes, so this may not misfire in practice -- but the method offers no protection if that assumption ever breaks (lock expiry, a bug in the lock store, some other caller), and it's inconsistent with the pattern the class itself documents and that `Tracker::Request` already follows correctly.

Capture the return value and guard the `notifiers.each` block on it, same shape as `Tracker::Request#transition_to_red`:

```ruby
  if state_store.transition_to_color(to_color)
    metrics_store.clear
    info = LightInfo.new(name: config.name)
    notifiers.each { |notifier| notifier.notify(info, from_color, to_color, nil) }
  end
```

pollychen-lab:
## Summary
- guard recovery probe cleanup and notification on `transition_to_color` returning true
- add regression coverage for duplicate recovery transitions that return false

Closes #748.

## Validation
- `git diff --check`
- `bundle exec rspec spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb` in `ruby:3.4` Docker container: 12 examples, 0 failures
- `bundle exec standardrb lib/stoplight/domain/tracker/recovery_probe.rb spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb` in `ruby:3.4` Docker container

### DIFF
--- a/lib/stoplight/domain/tracker/recovery_probe.rb
+++ b/lib/stoplight/domain/tracker/recovery_probe.rb
@@ -46,11 +46,12 @@ module Stoplight
             raise "recovery strategy returned unexpected color: #{recovery_result}"
           end
 
-          state_store.transition_to_color(to_color)
-          metrics_store.clear
-          info = LightInfo.new(name: config.name)
-          notifiers.each do |notifier|
-            notifier.notify(info, from_color, to_color, nil)
+          if state_store.transition_to_color(to_color)
+            metrics_store.clear
+            info = LightInfo.new(name: config.name)
+            notifiers.each do |notifier|
+              notifier.notify(info, from_color, to_color, nil)
+            end
           end
         end
       end
--- a/spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb
+++ b/spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb
@@ -17,7 +17,7 @@ RSpec.describe Stoplight::Domain::Tracker::RecoveryProbe do
 
       before do
         allow(traffic_recovery).to receive(:determine_color).with(config, metrics_after_probe).and_return(recover_to)
-        allow(state_store).to receive(:transition_to_color).with(transition_to)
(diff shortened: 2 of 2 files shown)


bolshakov:
@pollychen-lab hey there 👋 Thanks for the contribution! We decided to introduce a policy that "good first issue" tickets are reserved for human contributors, since they're designed specifically to help newcomers learn Stoplight's internals through a self-contained, narrow-scope first task. Since your contribution came in before this policy was in place, we'll review and accept it as-is -- no issue here.

Going forward, though, AI-assisted contributions won't be accepted for tickets in this category. Non-"good first issue" tickets don't have this restriction, so feel free to keep contributing there. Thanks again, and appreciate you taking the time!

bolshakov (approved):

bolshakov:
Thank you for your contribution 🙇

PR #755, opened 2026-07-12
Wire LightRegistered telemetry event

### LINKED ISSUE #752: Wire `LightRegistered` event

`LightRegistered`.

## What you're wiring up

`LightRegistered` (`lib/stoplight/domain/telemetry/light_registered.rb`,
`sig/stoplight/telemetry/light_registered.rbs`) fires once per `(process, system, light)` -- the
moment a given light's configuration first materializes. It's the "fleet inventory" event: a
subscriber can use it to build a live list of every light that exists, with what settings,
without waiting for that light to ever fail.

Read the sig's own doc comment closely on ordering: this event must be emitted **through the
light's own emitter, after the light is fully constructed**, and it must fire **before**
anything else that light does -- so on the bus it's guaranteed to precede every other event for
that light. Emitting it too early (before construction finishes) or from the wrong place breaks
that guarantee, so the placement below matters more than it looks like it should.

## Where to make the change

`Wiring::System#light`, inside the `lights.compute_if_absent` block  -- this block only runs the first time a given light
name is requested from this system; every later call returns the cached instance, so you get
"once per light" for free.

`built` (the `Domain::Light` `LightFactory#build` returns) doesn't currently expose its own
emitter to callers -- `LightFactory` holds it privately as `@emitter`. The simplest fix is
probably to have `LightFactory#build` itself emit `LightRegistered` right before returning,
since it already has both the emitter and the fully-assembled config in scope -- that keeps
`System#light` free of telemetry concerns entirely. 

## Payload fields

- `settings:` -- a `Domain::Telemetry::Settings` snapshot of the light's configuration. Look at
  `lib/stoplight/domain/telemetry/settings.rb` for what fields it expects, and compare against
  what `config`/`light_dsl` can supply at this point. Two things worth knowing before you start:

## Suggested approach

1. Trace `LightFactory#build` (`lib/stoplight/wiring/light_factory.rb`) to see what it already
   has in scope (`@emitter`, `config`) versus what `Settings` needs -- decide what's already
   there and what needs threading through.
2. Add spec coverage in `spec/unit/stoplight/wiring/system_spec.rb`, and/or
   `spec/unit/stoplight/wiring/light_factory_spec.rb`: the first `system.light("x")` call emits
   exactly one `LightRegistered`; a second call for the same name emits none; the `config_digest`
   used elsewhere matches what shows up in this event's settings.

## References

- `sig/stoplight/telemetry/light_registered.rbs`

Bortlesboat:
## Summary

- emit `LightRegistered` through each light's own emitter after construction
- include the normalized, serializable light configuration in `Telemetry::Settings`
- preserve once-per-light behavior through `System#light`'s existing cache-miss boundary

Closes #752.

## Validation

- focused registration spec: 1 example, 0 failures
- `bundle exec standardrb lib/stoplight/wiring/light_factory.rb spec/unit/stoplight/wiring/system_spec.rb`
- `bundle exec steep check lib/stoplight/wiring/light_factory.rb`
- `ruby -c lib/stoplight/wiring/light_factory.rb`
- `git diff --check`

The full Steep run still reports five existing errors outside this diff (`fail_safe.rb`, Redis key-space signatures, and `domain/telemetry.rb`). The complete `system_spec` also requires a local Redis server for its pre-existing Redis-tagged example; the new telemetry example passes in isolation.

### DIFF
--- a/lib/stoplight/wiring/light_factory.rb
+++ b/lib/stoplight/wiring/light_factory.rb
@@ -38,13 +38,17 @@ module Stoplight
       end
 
       def build
-        Stoplight::Domain::Light.new(
+        light = Stoplight::Domain::Light.new(
           @name,
           state_store:,
           green_run_strategy:,
           yellow_run_strategy:,
           red_run_strategy:
         )
+        @emitter.emit(Domain::Telemetry::LightRegistered) do
+          Domain::Telemetry::LightRegistered.new(settings: telemetry_settings)
+        end
+        light
       end
 
       def state_store = storage_set.state_store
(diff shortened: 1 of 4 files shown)


bolshakov:

PR #756, opened 2026-07-12
Wire LockChanged telemetry event

### LINKED ISSUE #751: Wire `LockChanged` event

`LockChanged`.

## What you're wiring up

`LockChanged` (`lib/stoplight/domain/telemetry/lock_changed.rb`,
`sig/stoplight/telemetry/lock_changed.rbs`) fires whenever a human (or a script) manually locks
or unlocks a light -- overriding whatever color the automatic circuit-breaker logic would pick.

Read the sig's own doc comment carefully: this event fires **even when the effective color
doesn't change** (e.g. locking an already-red light to red again still emits). The point is the
*lock action itself* being recorded, not a color transition -- don't add a "did the color
actually change" guard, that would be wrong here specifically (unlike the other events in this
set).

## Where to make the change

Two separate call paths both need this, since Stoplight has two ways to lock/unlock a light: `Domain::Light#lock` / `#unlock`

## Payload fields

- `from_color:` / `to_color:` -- read before/after `set_state`, or derive from the requested lock
  `state` (`State::LOCKED_RED` implies `Color::RED`, etc.) if reading "before" isn't cheap.
- `from_state:` / `to_state:` -- the light's locked state before and after the call (`State::LOCKED_RED`,
  `State::LOCKED_GREEN`, `State::UNLOCKED`).
- `source:` -- remove this field from the event. We decided not to add this information.

## References

- `sig/stoplight/telemetry/lock_changed.rbs`

Bortlesboat:
## Summary

- emit `LockChanged` for both manual lock and unlock operations
- capture state and effective color before and after the write
- emit even when relocking to the same effective color
- remove the abandoned `source` field from the event contract as requested

Closes #751.

## Validation

- `bundle exec rspec spec/unit/stoplight/domain/light_spec.rb` — 12 examples, 0 failures
- `bundle exec standardrb` on all changed Ruby/spec files
- `bundle exec steep check` on all changed library files — no type errors
- `git diff --check`

### DIFF
--- a/lib/stoplight/domain/light.rb
+++ b/lib/stoplight/domain/light.rb
@@ -11,12 +11,13 @@ module Stoplight
       attr_reader :red_run_strategy
       attr_reader :state_store
 
-      def initialize(name, green_run_strategy:, yellow_run_strategy:, red_run_strategy:, state_store:)
+      def initialize(name, green_run_strategy:, yellow_run_strategy:, red_run_strategy:, state_store:, emitter:)
         @name = name
         @green_run_strategy = green_run_strategy
         @yellow_run_strategy = yellow_run_strategy
         @red_run_strategy = red_run_strategy
         @state_store = state_store
+        @emitter = emitter
       end
 
       # Returns the current state of the light:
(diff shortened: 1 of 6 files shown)


bolshakov:

PR #757, opened 2026-07-13
feat: emit TrafficBreached telemetry

### LINKED ISSUE #743: Wire `TrafficBreached` event

The original design is PR #721 -- read its "Motivation" and "What this opens up" sections 
for the why. All the event types it proposed
actually wired up and firing today (PR #742, PR #741). 

This issue wires up one more: `TrafficBreached`.

## What you're wiring up

`TrafficBreached` (`lib/stoplight/domain/telemetry/traffic_breached.rb`,
`sig/stoplight/telemetry/traffic_breached.rbs`) fires when traffic control trips a light from
green to red -- the circuit just opened.

## Where to make the change

`Domain::Tracker::Request#transition_to_red` (`lib/stoplight/domain/tracker/request.rb:38-47`):

```ruby
def transition_to_red(exception, metrics:)
  if traffic_control.stop_traffic?(config, metrics)
    if state_store.transition_to_color(Color::RED)   # <- emit inside this branch only
      info = LightInfo.new(name: config.name)
      notifiers.each { |notifier| notifier.notify(info, Color::GREEN, Color::RED, exception) }
    end
  end
end
```

`state_store.transition_to_color(Color::RED)` returns `true` only the *first* time the light
actually flips to red -- if it's already red, it returns `false`. That's why the existing
`Notifier` call lives inside that `if`: without it, every subsequent failed request while
already red would notify again. Your emit call must live inside that same `if`, for the same
reason -- this is called the "dedup invariant" elsewhere in these docs.

## Payload fields

- `from_color:` -- always `Color::GREEN` here.
- `to_color:` -- always `Color::RED`.
- `policy:` -- a string the traffic-control policy uses to self-identify (e.g.
  `"consecutive_errors"`, `"error_rate"`). Check what `traffic_control` currently exposes for
  this -- if nothing does, you'll need to add a small reader.
- `failure:` -- `Domain::Telemetry::Failure.new(exception:, tracked: true)`. This call site is
  always reached from a recorded failure, so `exception` is always present.
- `metrics:` -- the `metrics` local a few lines up (`metrics_store.metrics_snapshot`), already
  computed by the caller.

## Suggested approach

1. `Tracker::Request` needs the emitter constructor-injected -- `LightFactory` already builds
   and passes a collaborator per strategy (see `run_recorder:` in
   `lib/stoplight/wiring/light_factory.rb`) if that shape fits, otherwise inject the emitter
   directly.
2. Add spec coverage in `spec/unit/stoplight/domain/tracker/request_spec.rb`: a red-transition
   emits exactly one `TrafficBreached` with the fields above; a request that doesn't trip the
   light (or trips it while already red) emits none.

## References

- `sig/stoplight/telemetry/traffic_breached.rbs`
- #721 -- original proposal
- #742, PR #741 -- `RunCompleted` reference implementation

Bortlesboat:
## Summary

Emit `TrafficBreached` when traffic control successfully transitions a light from green to red. The event carries the normalized traffic-control policy name, tracked failure, and a serializable snapshot of the metrics used for the breach decision.

The emit remains inside `transition_to_color`'s existing dedup guard, so a lost concurrent transition or a decision to continue traffic emits nothing.

Fixes #743.

## Verification

- `bundle exec rspec spec/unit/stoplight/domain` — 151 examples, 0 failures
- `bundle exec standardrb lib/stoplight/domain/tracker/request.rb lib/stoplight/wiring/light_factory.rb spec/unit/stoplight/domain/tracker/request_spec.rb`
- `bundle exec steep check` — changed files clean; five existing errors remain in Redis key-space, notifier fail-safe, and telemetry table code
- `git diff --check`

The broader wiring slice passed 27 non-Redis examples; its Redis-backed example requires a local Redis server and was not run successfully in this environment.

AI assistance disclosure: OpenAI Codex assisted with implementation; I reviewed the diff and verification results.

### DIFF
--- a/lib/stoplight/domain/tracker/request.rb
+++ b/lib/stoplight/domain/tracker/request.rb
@@ -10,12 +10,13 @@ module Stoplight
       #
       # @api private
       class Request
-        def initialize(traffic_control:, notifiers:, config:, metrics_store:, state_store:)
+        def initialize(traffic_control:, notifiers:, config:, metrics_store:, state_store:, emitter:)
           @traffic_control = traffic_control
           @notifiers = notifiers
           @config = config
           @metrics_store = metrics_store
           @state_store = state_store
+          @emitter = emitter
         end
 
         def record_failure(exception)
@@ -34,6 +35,7 @@ module Stoplight
         attr_reader :config
         attr_reader :metrics_store
         attr_reader :state_store
+        attr_reader :emitter
 
         def transition_to_red(exception, metrics:)
           if traffic_control.stop_traffic?(config, metrics)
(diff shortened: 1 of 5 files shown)


bolshakov:

PR #758, opened 2026-07-13
Allow per-call error tracking overrides

### LINKED ISSUE #738: Allow tracked_errors / skipped_errors to be overridden per-call on Light#run
Since systems became the composition root, a given `Light` name is registered once with one fixed `Config` (`tracked_errors`, `skipped_errors`, among others) — `Stoplight()` reject re-registering the same name with different settings. That’s the right constraint for things like `threshold` or `window_size`, which describe policy for the breaker as a whole.

But tracked_`errors`/`skipped_errors` are different: they describe which exceptions from this particular call site should count as a failure. In practice, the same light is often reused across call sites that each throw a different flavor of error. 

Forcing every call site to agree on one static error list, or forcing them into separate lights/registrations, works against that.

### Proposal

Add `tracked_errors`: and `skipped_errors`: keyword arguments to `Domain::Light#run`. 

When passed, they override the light-level config’s `tracked_errors`/`skipped_errors` for that invocation only. When omitted, behavior is unchanged (falls back to the light’s registered config, as today).

Where this touches code

- `Domain::ErrorTrackingPolicy` is currently constructed once per light from `config.tracked_errors`/`config.skipped_errors`, and shared by all three run strategies. It’s stateless/cheap to build, so a per-call policy is feasible.
- `Domain::Strategies::{Green,Yellow,Red}` `RunStrategy#execute` each close over `@error_tracking_policy` baked in at construction. They’d need to accept an optional override (or receive the policy per call instead of at construction).
- `Light#run` would need to build a one-off `ErrorTrackingPolicy` when either override is passed, and thread it through to `strategy.execute`.

### Open questions

1. Override semantics: does passing tracked_errors: at the call site replace the light’s list, or merge with it?
2. Does this need to interact with telemetry/notifiers, which also read tracked/skipped config off the light (`Domain::Telemetry::Settings`)? Or is the override strictly scoped to the traffic-control decision (trip to red or not) for that call?

oiahoon:
## Summary

- allow `Light#run` to replace `tracked_errors` or `skipped_errors` for one invocation while omitted values fall back to the registered configuration
- keep the default error-tracking policy as one shared value and derive per-call replacements through `ErrorTrackingPolicy#with`
- synchronize RBS and README documentation and cover replacement, partial fallback, empty lists, isolation, and all light colors

Closes #738

## Testing

- `bundle exec rspec` (624 examples)
- `bundle exec steep check`
- `bundle exec standardrb`
- full Cucumber matrix with Memory and Redis using both `Stoplight()` and `System#light` (62 scenarios per run; four expected global-configuration scenarios pending for `System#light`)
- `git diff --check`

### DIFF
--- a/features/stoplight/configuration.feature
+++ b/features/stoplight/configuration.feature
@@ -86,6 +86,33 @@ Feature: Stoplight Custom Configuration
     And 1 request is made
     Then the light color is green
 
+  Scenario: Per-call tracked errors replace the registered list and preserve registered skipped errors
+    Given a light configured with:
+      | Tracked Errors | KeyError       |
+      | Skipped Errors | Timeout::Error |
+      | Threshold      | 1              |
+    When the service starts failing with:
+      | Type | Timeout::Error |
+    And 1 request is made with:
+      | Tracked Errors | StandardError |
+    Then the light color is green
+    When the service starts failing with:
+      | Type | KeyError |
+    And 1 request is made with:
+      | Tracked Errors | Timeout::Error |
+    Then the light color is green
+
+  Scenario: Per-call skipped errors replace the registered list and preserve registered tracked errors
+    Given a light configured with:
+      | Tracked Errors | StandardError  |
+      | Skipped Errors | Timeout::Error |
+      | Threshold      | 1              |
+    When the service starts failing with:
+      | Type | Timeout::Error |
+    And 1 request is made with:
+      | Skipped Errors | KeyError |
+    Then the light color is red
+
   Scenario: System-level exceptions don't trigger circuit breaker
     Given a light configured with:
       | Threshold   | 1             |
--- a/lib/stoplight/domain/strategies/green_run_strategy.rb
+++ b/lib/stoplight/domain/strategies/green_run_strategy.rb
@@ -10,8 +10,7 @@ module Stoplight
       #
       # @api private
       class GreenRunStrategy
-        def initialize(error_tracking_policy:, request_tracker:, run_recorder:, clock:)
-          @error_tracking_policy = error_tracking_policy
+        def initialize(request_tracker:, run_recorder:, clock:)
           @request_tracker = request_tracker
           @run_recorder = run_recorder
           @clock = clock
@@ -24,7 +23,7 @@ module Stoplight
         # @yield The code block to execute.
         # @return The result of the code block if successful.
         # @raise re-raises the error if it is not tracked or no fallback is provided.
-        def execute(fallback, state_snapshot:, &code)
+        def execute(fallback, state_snapshot:, error_tracking_policy:, &code)
           started_at = capture_started_at
 
           begin
@@ -33,7 +32,7 @@ module Stoplight
 
             result
           rescue => error
-            if @error_tracking_policy.track?(error)
+            if error_tracking_policy.track?(error)
               record_error(error, duration_ms: duration_since(started_at), fallback_used: !fallback.nil?)
 
               if fallback
--- a/lib/stoplight/wiring/light_factory.rb
+++ b/lib/stoplight/wiring/light_factory.rb
@@ -44,7 +44,8 @@ module Stoplight
           green_run_strategy:,
           yellow_run_strategy:,
           red_run_strategy:,
-          lock_control:
+          lock_control:,
+          error_tracking_policy: @error_tracking_policy
         )
       end
 
@@ -103,7 +104,6 @@ module Stoplight
 
       def green_run_strategy
         Domain::Strategies::GreenRunStrategy.new(
-          error_tracking_policy: @error_tracking_policy,
           request_tracker:,
           clock:,
           run_recorder: run_recorder(Color::GREEN)
@@ -113,7 +113,6 @@ module Stoplight
       def yellow_run_strategy
         Domain::Strategies::YellowRunStrategy.new(
           name: @name,
-          error_tracking_policy: @error_tracking_policy,
           notifiers:,
           request_tracker: recovery_probe_tracker,
           state_store:,
--- a/spec/unit/stoplight/domain/light_spec.rb
+++ b/spec/unit/stoplight/domain/light_spec.rb
@@ -8,7 +8,8 @@ RSpec.describe Stoplight::Domain::Light do
       yellow_run_strategy:,
       red_run_strategy:,
       state_store:,
-      lock_control:
+      lock_control:,
+      error_tracking_policy:
     )
   end
   let(:config) { instance_double(Stoplight::Domain::Config) }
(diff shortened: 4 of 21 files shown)


bolshakov:
@oiahoon hey there 👋 Thanks for the contribution! We decided to introduce a policy that "good first issue" tickets are reserved for human contributors, since they're designed specifically to help newcomers learn Stoplight's internals through a self-contained, narrow-scope first task. Since your contribution came in before this policy was in place, we'll review and accept it as-is -- no issue here.

Going forward, though, AI-assisted contributions won't be accepted for tickets in this category. Non-"good first issue" tickets don't have this restriction, so feel free to keep contributing there. Thanks again, and appreciate you taking the time!

oiahoon:
> @oiahoon hey there 👋 Thanks for the contribution! We decided to introduce a policy that "good first issue" tickets are reserved for human contributors, since they're designed specifically to help newcomers learn Stoplight's internals through a self-contained, narrow-scope first task. Since your contribution came in before this policy was in place, we'll review and accept it as-is -- no issue here.
> 
> Going forward, though, AI-assisted contributions won't be accepted for tickets in this category. Non-"good first issue" tickets don't have this restriction, so feel free to keep contributing there. Thanks again, and appreciate you taking the time!

sounds good.

features/stoplight/configuration.feature
```diff
@@ -86,6 +86,33 @@ Feature: Stoplight Custom Configuration
...
+      | Skipped Errors | Timeout::Error |
+      | Threshold      | 1              |
+    When the service starts failing with:
+      | Type | Timeout::Error |
+    And 1 request is made with per-call error tracking:
+      | Tracked Errors | StandardError |
+    Then the light color is green
+    When the service starts failing with:
+      | Type | KeyError |
+    And 1 request is made with per-call error tracking:
```
bolshakov:
```suggestion
    And 1 request is made with:
```

This less specific phrasing would allow us add new parameters in the future without changing all call sites

lib/stoplight/wiring/light_factory.rb
```diff
@@ -43,7 +43,9 @@ def build
           state_store:,
           green_run_strategy:,
           yellow_run_strategy:,
-          red_run_strategy:
+          red_run_strategy:,
+          tracked_errors: config.tracked_errors,
+          skipped_errors: config.skipped_errors
```
bolshakov: `LightFactory` builds two independent `ErrorTrackingPolicy` equivalents from the same config: one real `ErrorTrackingPolicy` for the strategies (see initializer), and a second copy as raw `tracked_errors`/`skipped_errors` here into `Domain::Light.new` here.

Light stores those arrays and only touches them on a customized `#run` call, where it re-wraps them into a new `ErrorTrackingPolicy` to backfill whichever side of the override (tracked_errors/skipped_errors) wasn't passed.

They're guaranteed to agree today since both reads happen off the same config at construction time, but nothing enforces that structurally, and Light ends up holding a second, looser representation (raw arrays) of state the strategies already hold as a proper policy object.

This also means `#run` itself is two jobs -- dispatching to the right strategy for the current color *and* resolving how per-call overrides merge with the defaults, in an inline branch that duplicates the `strategy.execute(...)` call site across both paths. This violates single responsiblity principle, we try to follow. 

Suggestion: have Light hold the default `ErrorTrackingPolicy` itself (instead of the raw `tracked_errors`/`skipped_errors` arrays), and add an `ErrorTrackingPolicy#with(tracked: Undefined, skipped: Undefined)` that returns a new policy with only the given fields overridden, falling
back to self's values otherwise. Then:

```
# Light#run
error_tracking_policy = @error_tracking_policy.with(tracked: tracked_errors, skipped: skipped_errors)
strategy.execute(fallback, state_snapshot:, error_tracking_policy:, &code)
```

This collapses the two-branch if into one path, and Light no longer needs to reconstruct anything for the uncustomized case either: `@error_tracking_policy.with(...)` with both args Undefined just returns self.

This also isolates the overriding behaviour inside `ErrorTrackingPolicy#with`, where it can be unit-tested directly, without touching `Light#run`.

spec/unit/stoplight/domain/light_spec.rb
```diff
@@ -111,5 +116,17 @@
...
+    it "uses per-call overrides with registered values as partial fallbacks" do
+      expect(green_run_strategy).to receive(:execute).with(
+        fallback,
+        state_snapshot:,
+        error_tracking_policy: satisfy { |policy|
+          policy.track?(KeyError.new) && !policy.track?(Timeout::Error.new)
+        }
+      ) { "result" }
+
+      expect(light.run(fallback, tracked_errors: [KeyError], &code)).to eq("result")
```
bolshakov, line 103: The override here narrows `tracked_errors` from the default `[StandardError]` down to `[KeyError]`,  but `KeyError` is itself a `StandardError`, so `policy.track?(KeyError.new)` would be true whether the override actually took effect or was silently dropped. 

Maybe add a case that's only excluded because of the override, something tracked under the default but not under `[KeyError]`, e.g. `ArgumentError`:

```ruby
  error_tracking_policy: satisfy { |policy|
    policy.track?(KeyError.new) && !policy.track?(ArgumentError.new)
  }
```

`ArgumentError` is a `StandardError` (tracked by default) but isn't a `KeyError`, so this only passes if the override genuinely replaced the tracked list rather than being ignored.

bolshakov (commented):
I noticed a documentation gam. The "Custom Error Configuration" section in the README covers `tracked_errors:`/`skipped_errors:` at config time, but doesn't mention the new per-call override this PR adds. The YARD doc on `Light#run` itself already has a nice. Let's put that into the README too, since that's likely the first place people look.

features/stoplight/configuration.feature
oiahoon: Addressed in 55f33fd. Generalized the step wording to `request is made with:`.

lib/stoplight/wiring/light_factory.rb
oiahoon: Addressed in 55f33fd. Light now stores the shared default policy and delegates partial override resolution to `ErrorTrackingPolicy#with`, with direct unit coverage.

spec/unit/stoplight/domain/light_spec.rb
oiahoon: Addressed in 55f33fd. The unit expectation now also rejects `ArgumentError`, proving the tracked list was replaced rather than inherited.

oiahoon:
Addressed in 55f33fd. Documented per-call tracked/skipped overrides and their replacement/fallback semantics in the README.

lib/stoplight/domain/strategies/green_run_strategy.rb
```diff
@@ -24,7 +24,8 @@ def initialize(error_tracking_policy:, request_tracker:, run_recorder:, clock:)
         # @yield The code block to execute.
         # @return The result of the code block if successful.
         # @raise re-raises the error if it is not tracked or no fallback is provided.
-        def execute(fallback, state_snapshot:, &code)
+        def execute(fallback, state_snapshot:, error_tracking_policy: nil, &code)
```
bolshakov: One thing I missed earlier - this new argument is never nil and the @error_tracking_policy from the initializer is never taken into account. This means the one from initializer could be safely removed, right?

The same applies to yellow run strategy

oiahoon: Addressed in 9eaa98f. Green and Yellow no longer store a constructor policy; every strategy execution now requires the per-run policy, with Red accepting it only to preserve the common interface.

bolshakov (approved):
Amazing job 🚀

PR #761, opened 2026-07-13
feat: emit RecoveryFailed telemetry

### LINKED ISSUE #750: Wire `RecoveryFailed` event

`RecoveryFailed`.

## What you're wiring up

`RecoveryFailed` (`lib/stoplight/domain/telemetry/recovery_failed.rb`,
`sig/stoplight/telemetry/recovery_failed.rbs`) fires when the recovery policy gives up and sends
the light back to red (yellow -> red) -- the dependency is still unhealthy.

## Where to make the change

`Domain::Tracker::RecoveryProbe#recover`, the `to_color == Color::RED` branch. 

```ruby
transitioned = state_store.transition_to_color(to_color)
metrics_store.clear
notifiers.each { ... }
# emit RecoveryFailed only if transitioned && to_color == Color::RED
```

## Payload fields

- `from_color:` -- always `Color::YELLOW`.
- `to_color:` -- always `Color::RED`.
- `policy:` -- same self-identifying policy string as `RecoverySucceeded` needs.
- `failure:` -- `Telemetry::Failure` object constructed from exception caused tripping back to red.
- `metrics:` -- the `recovery_metrics` local already computed in `#recover`.

## Suggested approach

1. Shares the emitter-injection prerequisite with `RecoverySucceeded` -- if both are being
   worked on, don't inject the emitter into `Tracker::RecoveryProbe` twice.
2. Add spec coverage in `spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb`: a probe
   that flips yellow -> red emits exactly one `RecoveryFailed`; a probe that keeps the light
   yellow emits none.

## References

- `sig/stoplight/telemetry/recovery_failed.rbs`

w3lld1:
## Summary

- inject the existing per-light telemetry emitter into `RecoveryProbe`
- emit `RecoveryFailed` only after this process successfully transitions a light from yellow to red
- include the normalized recovery-policy name, triggering failure, and decision-point metrics in the event payload
- keep RBS signatures in sync and cover successful, lost-race, and still-yellow paths

## Validation

- `bundle exec rspec spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb` — 14 examples, 0 failures
- `STOPLIGHT_REDIS_URL=redis://127.0.0.1:6379/0 bundle exec rspec` — 613 examples, 0 failures
- `bundle exec standardrb lib/stoplight/domain/tracker/recovery_probe.rb lib/stoplight/wiring/light_factory.rb spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb`
- `bundle exec steep check --no-daemon --jobs=2`
- `git diff --check`

Fixes #750

### DIFF
--- a/lib/stoplight/domain/tracker/recovery_probe.rb
+++ b/lib/stoplight/domain/tracker/recovery_probe.rb
@@ -4,19 +4,20 @@ module Stoplight
   module Domain
     module Tracker
       class RecoveryProbe
-        def initialize(traffic_recovery:, notifiers:, config:, metrics_store:, state_store:)
+        def initialize(traffic_recovery:, notifiers:, config:, metrics_store:, state_store:, emitter:)
           @traffic_recovery = traffic_recovery
           @notifiers = notifiers
           @config = config
           @metrics_store = metrics_store
           @state_store = state_store
+          @emitter = emitter
         end
 
         # @param exception [Exception]
         def record_failure(exception)
           metrics_store.record_failure(exception)
 
-          recover
+          recover(exception)
         end
 
         def record_success
@@ -32,8 +33,9 @@ module Stoplight
         attr_reader :config
         attr_reader :metrics_store
         attr_reader :state_store
+        attr_reader :emitter
 
-        def recover
+        def recover(exception = nil)
           recovery_metrics = metrics_store.metrics_snapshot
           recovery_result = traffic_recovery.determine_color(config, recovery_metrics)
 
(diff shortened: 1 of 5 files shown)


bolshakov:
Hey there 👋 Thanks for the contribution! Unfortunately we can't accept it because it appears to be AI-generated. Nothing against AI -- but "good first issue" tickets are designed specifically to bring in new human contributors. We write them self-contained, descriptive, and narrow in scope so newcomers can orient in the codebase and make a first contribution. 

Over time that's meant to help them learn Stoplight's internals and take on bigger work. AI contributions don't get that learning benefit, so we reserve this category for people, not tools.

We genuinely appreciate the effort here. For next time, the pinned comment covers this and walks through how to make a first contribution.

Issue #788, opened 2026-07-14
`Generic#notify` should use `_LightInfo`, not `Domain::Config`


bolshakov:
## Summary

`Generic#notify` implements `_StateTransitionNotifier`, whose port types the first argument as
`_LightInfo`. Both the sigs and one real caller still say `Domain::Config` instead. Fix all three so the
type and the code agree with the port.

## What to change

1. `sig/_private/stoplight/infrastructure/notifier/generic.rbs:12` — `notify`'s first argument type.
2. `sig/stoplight.rbs:29` — the public `notification_formatter` alias, same argument. Must change
   together with #1: `Generic#notify` forwards `light` into `formatter.call(light, ...)`
   (`lib/stoplight/infrastructure/notifier/generic.rb:63`), which is typed through this alias.
3. `lib/stoplight/domain/strategies/yellow_run_strategy.rb:123-133` (`enter_recovery`) — the one runtime
   caller still passing `@config` to `notify`. There's a commented-out `FIXME` right there sketching the
   fix. `@config` stays needed elsewhere in the class (`cool_off_time` in `with_recovery_lock`), so only
   this call site changes.

## Acceptance criteria

- [ ] All three edits applied.
- [ ] Spec tightened and confirmed red before the code fix, green after.
- [ ] `bundle exec steep check`, `bundle exec standardrb`, `bundle exec rspec` all pass.

PR #792, opened 2026-07-15
Use Domain::LightInfo in Generic#notify instead of Domain::Config

shkrt:
Addresses https://github.com/bolshakov/stoplight/issues/788

### DIFF
--- a/lib/stoplight/domain/strategies/yellow_run_strategy.rb
+++ b/lib/stoplight/domain/strategies/yellow_run_strategy.rb
@@ -125,10 +125,9 @@ module Stoplight
 
           state_store.transition_to_color(Color::YELLOW)
           metrics_store.clear
-          # FIXME: use light config instead of @config
-          # light_info = LightInfo.new(name: @name)
+          light_info = LightInfo.new(name: @name)
           notifiers.each do |notifier|
-            notifier.notify(@config, Color::RED, Color::YELLOW, nil)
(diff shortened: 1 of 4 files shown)


spec/unit/stoplight/domain/run_strategies/yellow_run_strategy_spec.rb
```diff
@@ -286,7 +286,8 @@
       it "notifies if able to transition to YELLO" do
         expect(metrics_store).to receive(:clear)
         expect(state_store).to receive(:transition_to_color).with(Stoplight::Color::YELLOW)
-        expect(notifier).to receive(:notify).with(have_attributes(name:), Stoplight::Color::RED, Stoplight::Color::YELLOW, nil)
+        expect(notifier).to receive(:notify)
```
bolshakov: Let's use the previous behavioral form here:

```ruby
          expect(notifier).to receive(:notify).with(have_attributes(name:), Stoplight::Color::GREEN, Stoplight::Color::RED, exception)
```

it asserts the `name` - what consumers actually care about and steep still enforces the contract on the CI.

bolshakov (commented):
Please also update also `@param config` here https://github.com/bolshakov/stoplight/blob/develop/sig/stoplight/ports/state_transition_notifier.rbs#L5

The comments itself is fine, just change it to `@param light`  or something like this

bolshakov (approved):

bolshakov:
Thank you Ruslan!

PR #802, opened 2026-07-15
Remove unused config reader from GreenRunStrategy

### LINKED ISSUE #784: Dead code / RBS mismatch: unused `config` reader in `GreenRunStrategy`
**Priority:** Low

## Summary
`GreenRunStrategy` declares `attr_reader :config`, but `@config` is never assigned, so the reader is unused dead
code that would silently return `nil` if ever invoked, and has no corresponding entry in the RBS sig.

## Location
`lib/stoplight/domain/strategies/green_run_strategy.rb:48`

## Details
`lib/stoplight/domain/strategies/green_run_strategy.rb:48` declares `attr_reader :config`, but `@config` is never
assigned (`initialize` only sets `@error_tracking_policy` and `@request_tracker`), and `green_run_strategy.rbs` has
no corresponding `config` reader. It is unused and would silently return `nil` if ever invoked.

## Suggested fix
Remove the stray `attr_reader :config`.

johnsonsirv:
## Summary
- Remove the stray private `attr_reader :config` from `GreenRunStrategy`.
- `@config` is never assigned in `initialize`, so the reader is unused dead code and is not present in the RBS sig.
- Fixes #784

## Test plan
- [X] Confirm `attr_reader :config` is gone from `lib/stoplight/domain/strategies/green_run_strategy.rb`
- [x] Confirm `attr_reader :request_tracker` remains (still used)
- [x] `bundle exec rspec spec/unit/stoplight/domain/run_strategies/green_run_strategy_spec.rb`

### DIFF
--- a/lib/stoplight/domain/strategies/green_run_strategy.rb
+++ b/lib/stoplight/domain/strategies/green_run_strategy.rb
@@ -51,7 +51,6 @@ module Stoplight
 
         private
 
-        attr_reader :config
         attr_reader :request_tracker
 
         def capture_started_at


bolshakov (approved):

bolshakov:
Thank you for the confirmation!

PR #804, opened 2026-07-15
Guard ErrorRate#check_compatibility against non-numeric threshold

### LINKED ISSUE #771: ErrorRate#check_compatibility raises NoMethodError on non-numeric threshold (no type guard at all)
## Summary

`check_compatibility` in `ErrorRate` has no type/nil guard for `threshold` whatsoever — only a range check
(`threshold < 0 || threshold > 1`). A `nil` or otherwise non-numeric `threshold` raises a raw `NoMethodError`
instead of returning the intended `CompatibilityResult.incompatible`. This is the same class of defect as the
`ConsecutiveErrors` issue, but worse, since there is no existing type check to merely reorder.

## Details

Reproduced: `Stoplight("t2", traffic_control: :error_rate, threshold: nil, window_size: 60)` raises
`NoMethodError: undefined method '<' for nil`.

## Suggested fix

Add a guard such as `!config.threshold.is_a?(Numeric)` before the `threshold < 0 || threshold > 1` range
comparison, returning `CompatibilityResult.incompatible`.

johnsonsirv:
## Summary
- Add a `Numeric` type guard in `ErrorRate#check_compatibility` before the `0..1` range check.
- Non-numeric / `nil` thresholds now return `CompatibilityResult.incompatible` instead of raising `NoMethodError`.
- Fixes #771

## Test plan
- [x] Unit specs cover `threshold: nil` and a non-numeric value (e.g. `"0.5"`), expecting `` `threshold` should be a number ``
- [x] Existing out-of-range cases still return `` `threshold` should be between 0 and 1 ``
- [x] `bundle exec rspec spec/unit/stoplight/domain/traffic_control/error_rate_spec.rb`
- [x] Smoke: `Stoplight("t2", traffic_control: :error_rate, threshold: nil, window_size: 60)` raises `ConfigurationError`, not `NoMethodError`

### DIFF
--- a/lib/stoplight/domain/traffic_control/error_rate.rb
+++ b/lib/stoplight/domain/traffic_control/error_rate.rb
@@ -26,6 +26,8 @@ module Stoplight
         def check_compatibility(config)
           if config.window_size.nil?
             CompatibilityResult.incompatible("`window_size` should be set")
+          elsif !config.threshold.is_a?(Numeric)
+            CompatibilityResult.incompatible("`threshold` should be a number")
           elsif config.threshold < 0 || config.threshold > 1
             CompatibilityResult.incompatible("`threshold` should be between 0 and 1")
           else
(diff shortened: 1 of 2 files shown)


bolshakov (approved):

bolshakov:
🔥

PR #806, opened 2026-07-16
Reorder ConsecutiveErrors#check_compatibility type check before range

### LINKED ISSUE #769: ConsecutiveErrors#check_compatibility raises NoMethodError on non-numeric threshold instead of returning incompatible result
## Summary
`check_compatibility` in `ConsecutiveErrors` checks `config.threshold <= 0` before checking
`config.threshold.is_a?(Integer)`. Because the numeric comparison runs first, a `nil` (or any other non-numeric)
`threshold` raises a raw `NoMethodError` instead of returning the intended `CompatibilityResult.incompatible` /
surfacing as `Stoplight::Error::ConfigurationError`.

## Details

Reproduced: `Stoplight("t1", threshold: nil)` raises `NoMethodError: undefined method '<=' for nil`. Nil is 
invalid value here, but it should not through `NoMethodError`.

## Suggested fix

Reorder the guards so `!config.threshold.is_a?(Integer)` is checked before the `threshold <= 0` range comparison,
so a `nil`/non-numeric `threshold` returns `CompatibilityResult.incompatible` instead of raising `NoMethodError`.

johnsonsirv:
## Summary
- Run `threshold.is_a?(Integer)` before `threshold <= 0` in `ConsecutiveErrors#check_compatibility`.
- `nil` / non-Integer thresholds now return `CompatibilityResult.incompatible` instead of raising `NoMethodError` / `ArgumentError`.
- Fixes #769

## Test plan
- [x] Specs cover `threshold: nil` and a non-Integer (e.g. `"3"`), expecting `` `threshold` should be an integer ``
- [x] Existing cases still pass (`0` => bigger than 0; `14.87` => should be an integer)
- [x] `bundle exec rspec spec/unit/stoplight/domain/traffic_control/consecutive_errors_spec.rb`
- [x] Smoke: `Stoplight("t1", threshold: nil)` raises `ConfigurationError`, not `NoMethodError`

### DIFF
--- a/lib/stoplight/domain/traffic_control/consecutive_errors.rb
+++ b/lib/stoplight/domain/traffic_control/consecutive_errors.rb
@@ -28,10 +28,10 @@ module Stoplight
       # @api private
       class ConsecutiveErrors
         def check_compatibility(config)
-          if config.threshold <= 0
-            CompatibilityResult.incompatible("`threshold` should be bigger than 0")
-          elsif !config.threshold.is_a?(Integer)
+          if !config.threshold.is_a?(Integer)
             CompatibilityResult.incompatible("`threshold` should be an integer")
+          elsif config.threshold <= 0
+            CompatibilityResult.incompatible("`threshold` should be bigger than 0")
           else
             CompatibilityResult.compatible
           end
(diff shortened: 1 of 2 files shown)


bolshakov (approved):
That's amazing 🚀

PR #810, opened 2026-07-16
fix: validate telemetry subscribe filter before mutating state

### LINKED ISSUE #768: Invalid filter in `Telemetry#subscribe` permanently poisons the event bus
## Summary
`Telemetry#subscribe` pushes `[subscription, filter, handler]` onto `@subscriptions` before calling `rebuild`,
which does `event_class <= filter` with no validation that `filter` is `nil` or a `Module`/class. A single bad
call (e.g. `telemetry.subscribe("typo") { ... }`) raises `TypeError` and leaves the bus permanently unusable.

## Details

`Module#<=` raises `TypeError` ("compared with non class/module") when compared against anything that isn't a
`Module`/`Class`. Consequences of a single bad call such as `telemetry.subscribe("typo") { ... }`:

1. The `TypeError` propagates out of `subscribe` from inside `@mutex.synchronize`, so the caller never reaches
   the `subscription` return value - the token is never returned, so the poisoned entry can never be targeted
   by `unsubscribe`.
2. The bad tuple is left in `@subscriptions` because there is no rollback on the `rebuild` failure.
3. Every subsequent `subscribe` or `unsubscribe` call — from completely unrelated, correct call sites — now also
   invokes `rebuild`, which iterates `@subscriptions` and hits the same poisoned tuple again, raising `TypeError`
   again. The telemetry bus for that light/system becomes permanently unusable for the life of the process, with
   no public API path to remove the poisoned entry.

This is reachable directly from end-user code: `Domain::Telemetry` is wired unwrapped as the public consumer
surface, so nothing upstream validates `filter` before it reaches `subscribe`. RBS constrains the type only at 
Steep-checked call sites, not at runtime, so this is a real runtime hazard, not just a hypothetical type violation.

## Suggested fix

Validate `filter` before mutating shared state (rather than rescuing/rolling back after a mid-mutation failure),
e.g. raise `ArgumentError` up front unless `filter.nil? || filter.is_a?(Module)`.

ankuanku1:
Fixes [#768](https://github.com/bolshakov/stoplight/issues/768)

Hello! This PR adds a check in `Telemetry#subscribe` so that if a user passes in a filter that isn't a `Module` (or `nil`), it raises an `ArgumentError `instead of crashing later with a `TypeError`.

Also added some tests in `telemetry_spec.rb`, invalid filter raises the right error, it doesn't get stuck in the internal list, and normal `subscribe/publish/unsubscribe` still work after a bad call.

Let me know if I'm missing anything!

### DIFF
--- a/lib/stoplight/domain/telemetry.rb
+++ b/lib/stoplight/domain/telemetry.rb
@@ -28,6 +28,7 @@ module Stoplight
 
       def subscribe(filter = nil, &handler)
         raise ArgumentError, "nothing to subscribe. Please, pass a block into `Telemetry#subscribe`" unless handler
+        raise ArgumentError, "filter must be nil or a Module, got #{filter.inspect}" unless filter.nil? || filter.is_a?(Module)
 
         subscription = Subscription.new
         @mutex.synchronize do
--- a/spec/unit/stoplight/domain/telemetry_spec.rb
+++ b/spec/unit/stoplight/domain/telemetry_spec.rb
@@ -107,6 +107,36 @@ RSpec.describe Stoplight::Domain::Telemetry do
       end
     end
 
+    context "with an invalid filter" do
+      subject(:bus) { described_class.new(error_notifier:, max_subscriptions: 1) }
+
+      it "raises ArgumentError instead of comparing it against event classes" do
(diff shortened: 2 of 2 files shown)


bolshakov (approved):

bolshakov:
Thank you for your contribution Anssi 🚀

PR #825, opened 2026-07-17
fix: bucket alignment at hour boundaries

### LINKED ISSUE #778: `WindowMetrics`'s write-bucket and read-window-bucket formulas disagree at exact hour boundaries
## Summary

`bucket_key` (used to choose which bucket a new event is written into) and `buckets_for_window` (used to choose
which buckets a snapshot read scans) round a timestamp to its containing hour differently. They agree everywhere
except when a timestamp lands exactly on a bucket boundary (a multiple of `bucket_size`), where `buckets_for_window`
stops one bucket short. An event written into that bucket is then silently excluded from any snapshot computed at
that boundary instant.

## Location

`lib/stoplight/infrastructure/redis/storage/window_metrics.rb:121-123` (`#bucket_key`, used via `#successes_key`/
`#errors_key` on the write path), `lib/stoplight/infrastructure/redis/storage/window_metrics.rb:131-143`
(`#buckets_for_window`, used on the read path in `#metrics_snapshot`)

## Details

```ruby
def bucket_key(metric:, time:)
  key_space.key(:window_metrics, metric, (time.to_i / bucket_size) * bucket_size)
end

def buckets_for_window(metric:, window_end:)
  window_end_ts = window_end.to_i
  window_start_ts = window_end_ts - @window_size
  start_bucket = (window_start_ts / bucket_size) * bucket_size
  end_bucket = ((window_end_ts - 1) / bucket_size) * bucket_size   # <-- the `-1`
  (start_bucket..end_bucket).step(bucket_size).map { |bucket_start| bucket_key(metric:, time: bucket_start) }
end
```

`bucket_key` floors `time` to its containing hour with no adjustment. `buckets_for_window`'s `end_bucket` floors
`window_end - 1` instead of `window_end` when picking the last bucket to scan. These formulas agree for every
`time`/`window_end` except when the value is an exact multiple of `bucket_size` (3600), where `buckets_for_window`
computes the *previous* hour's bucket instead of the current one.

Concretely, with `bucket_size = 3600`:
- `bucket_key(time: 1696154400)` -> `1696154400` (that hour's own bucket) - this is the bucket
  `record_success`/`record_failure` writes the new event into via `successes_key`/`errors_key`.
- `buckets_for_window(window_end: 1696154400)`'s `end_bucket` -> `((1696154400 - 1) / 3600) * 3600 = 1696150800` -
  the **previous** hour.

So an event written at exactly `T = 1696154400` lands in bucket `1696154400`, but a `metrics_snapshot` computed with
`window_end_ts` equal to that same instant never enumerates that bucket key at all - `metrics_snapshot`'s Lua
(`window_metrics/_metrics_snapshot.lua`) only runs `ZCOUNT` over whichever bucket keys got passed in via `KEYS`, so
an event in a bucket that was never enumerated is not undercounted by range - it's excluded outright.

This is a narrow window (only triggers when a timestamp is an exact multiple of 3600), which is likely why it
hasn't surfaced as a flaky spec. 

## Suggested fix

Reproduced the issue in a test! Either `WindowMetrics` unit test with mocked clock or, even better, a proportry 
test (see `spec/properties/`) that verifies that `buckets_for_window` and `bucket_key` always alighned.

Extract a single shared end-bucket-alignment helper and have both the write path (`successes_key`/`errors_key` via
`bucket_key`) and the read path (`buckets_for_window`'s `end_bucket`) call it, rather than maintaining two
independent flooring formulas that must be kept in sync by hand.

ankuanku1:
Fixes #778 

This PR extracts a shared `bucket_start_for` helper and has both `bucket_key` and `buckets_for_window` use it, removing the mismatched `-1` in `buckets_for_window`'s end bucket calculation so the two stay aligned.

Also added a property test in `spec/properties/window_metrics_bucket_alignment_spec.rb` that checks `bucket_key` and `buckets_for_window` always agree on which bucket a timestamp belongs to, and fixed two existing tests in `window_metrics_spec.rb` that were asserting the old behavior at an hour boundary.

Let me know if it needs any changes!

### DIFF
--- a/lib/stoplight/infrastructure/redis/storage/window_metrics.rb
+++ b/lib/stoplight/infrastructure/redis/storage/window_metrics.rb
@@ -119,7 +119,7 @@ module Stoplight
           # @param time [Time, Numeric] The time for which to generate the key.
           # @return [String] The generated Redis key.
           def bucket_key(metric:, time:)
-            key_space.key(:window_metrics, metric, (time.to_i / bucket_size) * bucket_size)
+            key_space.key(:window_metrics, metric, bucket_start_for(time))
           end
 
           # Retrieves the list of Redis bucket keys required to cover a specific time window.
@@ -133,10 +133,10 @@ module Stoplight
             window_start_ts = window_end_ts - @window_size
 
             # Find bucket timestamps that contain any part of the window
-            start_bucket = (window_start_ts / bucket_size) * bucket_size
+            start_bucket = bucket_start_for(window_start_ts)
 
-            # End bucket is the last bucket that contains data within our window
-            end_bucket = ((window_end_ts - 1) / bucket_size) * bucket_size
+            # End bucket is the bucket that window_end itself falls into, so it's always included
+            end_bucket = bucket_start_for(window_end_ts)
 
             (start_bucket..end_bucket).step(bucket_size).map do |bucket_start|
               bucket_key(metric: metric, time: bucket_start)
@@ -150,6 +150,8 @@ module Stoplight
           attr_reader :metrics_key
           attr_reader :clock
           attr_reader :key_space
+
+          def bucket_start_for(time) = (time.to_i / bucket_size) * bucket_size
 
           def bucket_size = 3600 # 1 hour
           def bucket_ttl = @window_size + bucket_size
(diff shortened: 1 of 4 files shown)


bolshakov (approved):
Thank you for this contribution! Amazing job 👏 

chuck-norris

PR #834, opened 2026-07-20
Reject telemetry subscribe filters that match no event class

### LINKED ISSUE #767: `Telemetry#subscribe` accepts filters that never match any event class, producing silent dead subscriptions
## Summary

Even after fixing the `TypeError`-poisoning defect, a syntactically valid but semantically wrong filter (e.g.
some unrelated `Module`/`Class` that never matches any `EVENT_CLASSES` member) silently produces a subscription
that never fires, with no error.

## Details

Not a crash, but a trap for typos (e.g. subscribing to the wrong class). An operator wiring up telemetry has
no way to discover their subscription is dead until they notice the absence of events they expected. This is an
unobservable-failure pattern from a reliability standpoint as well as a correctness gap.

## Suggested fix

Validate that `filter` is `nil`, one of `EVENT_CLASSES`, or `StateTransitioned` (or otherwise actually matched by
at least one), and raise `ArgumentError` otherwise — mirroring the existing "nothing to subscribe" guard at
`telemetry.rb:30` — so a typo'd filter fails loudly instead of silently registering a dead subscription.

SAY-5:
Closes #767.

`Telemetry#subscribe` already rejects non-Module filters, but a valid Module that isn't one of the event classes (or an ancestor of one) still passes and registers a subscription that can never fire. This adds a guard that raises `ArgumentError` when the filter matches none of `EVENT_CLASSES`, using the same `event_class <= filter` check `rebuild` uses, so a typo'd filter fails at subscription time instead of silently going dead. `nil` (firehose) and `StateTransitioned` still work.

### DIFF
--- a/lib/stoplight/domain/telemetry.rb
+++ b/lib/stoplight/domain/telemetry.rb
@@ -29,6 +29,9 @@ module Stoplight
       def subscribe(filter = nil, &handler)
         raise ArgumentError, "nothing to subscribe. Please, pass a block into `Telemetry#subscribe`" unless handler
         raise ArgumentError, "filter must be nil or a Module, got #{filter.inspect}" unless filter.nil? || filter.is_a?(Module)
+        unless filter.nil? || EVENT_CLASSES.any? { |event_class| event_class <= filter }
+          raise ArgumentError, "filter #{filter.inspect} does not match any telemetry event class"
+        end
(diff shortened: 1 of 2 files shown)


bolshakov (approved):
Great job! Thank you 🚀

PR #836, opened 2026-07-21
fix: scope the rescue to the guarded call in run strategies

### LINKED ISSUE #766: Rescue scope in GreenRunStrategy#execute and YellowRunStrategy#execute swallows the strategy's own post-success bookkeeping
## Summary

The `begin`/`rescue` region in `GreenRunStrategy#execute` and `YellowRunStrategy#execute` covers not just
the user's guarded `code.call` but also the strategy's own post-success bookkeeping calls. If bookkeeping raises,
it is misreported as a failure of the caller's code, corrupting metrics and discarding a successful result.

## Details

```ruby 
def execute(fallback, state_snapshot:, &code)
  started_at = capture_started_at

  begin
    result = code.call                                      # <- user's code
    record_success(duration_ms: duration_since(started_at)) # <- stoplight code, both protected by begin/rescue 

    result
  rescue => error
    # ....
  end
end
```

In `GreenRunStrategy#execute`, the implicit `begin` region covers both `code.call` and the following
`record_success` call. In `YellowRunStrategy#execute`, the block covers `enter_recovery(state_snapshot)` and
`record_recovery_probe_success`, not just `code.call`. If any of those bookkeeping calls raises, the
exception is indistinguishable from a failure of the caller's own code: it is passed to
`error_tracking_policy.track?`, recorded via `record_error`/`record_recovery_probe_failure` (corrupting the
success/failure metrics that drive tripping/recovery even though the guarded code actually succeeded), and, if a
fallback exists, the fallback is invoked with this synthetic error while the real, successful `result` is silently
discarded.

Not currently covered by specs (`enter_recovery` is stubbed out in `yellow_run_strategy_spec.rb:31`, so this path
is untested).

## Suggested fix

Scope the `rescue` to `code.call` only, then run `record_success` (and, for yellow, `enter_recovery` /
`record_recovery_probe_success`) outside the guarded region. Something like this:

```ruby 
begin
  code.call
rescue => error
  # no changes here 
else
  record_success(...)
end
```

Add a unit test that makes a bookkeeping collaborator (`request_tracker`/`state_store`/`metrics_store`) raise on the 
success path, to pin down the expected behavior once fixed.

snowyukitty:
Fixes #766.

## Problem

`GreenRunStrategy#execute` and `YellowRunStrategy#execute` wrapped the strategy's own post-success bookkeeping inside the same `begin`/`rescue` region as the user's guarded `code.call`. When a bookkeeping call raised on the success path, the exception was indistinguishable from a failure of the caller's code:

- it was passed to `error_tracking_policy.track?`,
- recorded via `record_error` / `record_recovery_probe_failure`, corrupting the success/failure metrics that drive tripping and recovery, and
- when a fallback was configured, the fallback was invoked with this synthetic error while the real, successful result was silently discarded.

Concretely:

- **Green** — the implicit `begin` covered `code.call` **and** the following `record_success`.
- **Yellow** — the recovery block covered `enter_recovery` (before the call) and `record_recovery_probe_success` (after it), not just `code.call`.

## Fix

Scope the `rescue` to `code.call` only and run the success bookkeeping in an `else` clause; for yellow, run `enter_recovery` ahead of the guarded region as well. A bookkeeping/recovery failure on a successful call now surfaces as itself instead of being reclassified as a tracked run failure. The recovery lock is still released by the existing `ensure` in `with_recovery_lock`. No method signatures or RBS changed.

## Tests

Added unit regressions that make a bookkeeping collaborator raise on the success path (as suggested in the issue):

- **green** — a raising `record_success` must surface the error, and must not consult `track?`, call `record_failure`, or invoke the fallback.
- **yellow** — the same for a raising `record_recovery_probe_success` and for a raising `enter_recovery`; both also assert the recovery lock is still released.

Each fails on the current code (the strategy swallows the error and returns the fallback value) and passes with the fix. The domain strategy files stay at 100% line coverage.

Validated with `bundle exec rspec` (712 examples, 0 failures), `STOPLIGHT_DATA_STORE=Memory` and `STOPLIGHT_DATA_STORE=Redis bundle exec cucumber` (62 scenarios each), `bundle exec standardrb`, and `bundle exec steep check` — all green.

### DIFF
--- a/lib/stoplight/domain/strategies/green_run_strategy.rb
+++ b/lib/stoplight/domain/strategies/green_run_strategy.rb
@@ -28,9 +28,6 @@ module Stoplight
 
           begin
             result = code.call
-            record_success(duration_ms: duration_since(started_at))
-
-            result
           rescue => error
             if error_tracking_policy.track?(error)
               record_error(error, duration_ms: duration_since(started_at), fallback_used: !fallback.nil?)
@@ -45,6 +42,9 @@ module Stoplight
               record_success(duration_ms: duration_since(started_at), error: error)
               raise
             end
+          else
+            record_success(duration_ms: duration_since(started_at))
+            result
           end
         end
 
--- a/lib/stoplight/domain/strategies/yellow_run_strategy.rb
+++ b/lib/stoplight/domain/strategies/yellow_run_strategy.rb
@@ -49,21 +49,24 @@ module Stoplight
           with_recovery_lock(fallback:, state_snapshot:) do |started_at|
             enter_recovery(state_snapshot)
 
-            result = code.call
-            record_recovery_probe_success(duration_ms: duration_since(started_at))
-            result
-          rescue => error
-            if error_tracking_policy.track?(error)
-              record_recovery_probe_failure(error, duration_ms: duration_since(started_at), fallback_used: !fallback.nil?)
+            begin
+              result = code.call
+            rescue => error
+              if error_tracking_policy.track?(error)
+                record_recovery_probe_failure(error, duration_ms: duration_since(started_at), fallback_used: !fallback.nil?)
 
-              if fallback
-                fallback.call(error)
+                if fallback
+                  fallback.call(error)
+                else
+                  raise
+                end
               else
+                record_recovery_probe_success(duration_ms: duration_since(started_at), error:)
                 raise
               end
             else
-              record_recovery_probe_success(duration_ms: duration_since(started_at), error:)
-              raise
+              record_recovery_probe_success(duration_ms: duration_since(started_at))
+              result
             end
           end
         end
--- a/spec/unit/stoplight/domain/run_strategies/green_run_strategy_spec.rb
+++ b/spec/unit/stoplight/domain/run_strategies/green_run_strategy_spec.rb
@@ -138,4 +138,27 @@ RSpec.describe Stoplight::Domain::Strategies::GreenRunStrategy do
       end
     end
   end
+
+  context "when success bookkeeping raises" do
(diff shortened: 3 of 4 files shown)


bolshakov:
Thanks for the thorough fix here! I appreciate that you added tests to `enter_recovery` in yellow..

A few small things I'd like cleaned up before merging:

1. Drop the "Regression for #766" comment blocks in both spec. They restate the issue history rather than an invariant
the code owns. That kind of comments age now very well and rather liability. 

2. Fold the "still releases the recovery lock" tests into their sibling test in the two new yellow contexts, rather than separate it blocks - that matches how the rest of the file already handles it (e.g. the existing failure-branch tests assert lock release inline with the primary assertion). Two standalone examples for this is more than we need.

3. The `allow(request_tracker).to receive(:record_success)` stub in the "when entering recovery raises" context is dead -` enter_recovery` raises before `code.call` (and therefore before  `record_recovery_probe_success`) is ever reached, so that line can go.

None of this is a correctness concern -- the fix itself is right. Just want to tidy up the tests before we merge.

snowyukitty:
Thanks for the review! Addressed all three:

1. Dropped the `Regression for #766` comment blocks in both specs.
2. Folded the lock-release assertions into the primary examples in the two new yellow contexts, matching how the existing failure-branch tests assert release inline.
3. Removed the dead `record_success` stub from the "when entering recovery raises" context — `enter_recovery` raises before the probe is ever reached.

Each test still fails on the pre-fix code and passes with the fix; full suite, `standardrb`, and `steep` are green.

bolshakov (approved):

bolshakov:
Thank you, great job!

bolshakov (approved):

PR #840, opened 2026-07-23
Add eql? and hash to Domain::Failure for value equality

### LINKED ISSUE #781: Domain::Failure defines == without eql?/hash, breaking Hash/Set value-equality contract
## Summary
`Domain::Failure` overrides value equality via `==`, but leaves `eql?`/`hash` at `Object`'s identity-based
defaults, so two value-equal `Failure` instances will not collide as `Hash` keys or dedupe via `Set`/
`Array#uniq`.

## Location
`lib/stoplight/domain/failure.rb:36-41`

## Details
Value equality is overridden via `==`, but `eql?`/`hash` are left at `Object`'s identity-based defaults, so two
value-equal `Failure` instances will not collide as `Hash` keys or dedupe via `Set`/`Array#uniq`. No current call
site relies on this, so it is not an active bug, but it is an inconsistent equality contract worth closing while
the class is this small.

## Suggested fix

`alias_method :eql?, :==` plus `def hash = [error_class, error_message, time].hash`.

johnsonsirv:
## Summary
- Add `eql?` and `hash` on `Domain::Failure` so they match field-based `==`.
- Value-equal failures now collide as Hash keys and dedupe via Set / `Array#uniq`.
- Fixes #781

## Test plan
- [x] Specs cover `eql?` and equal `hash` when `==` is true
- [x] Specs show Hash key collision / Set or `uniq` size of 1 for value-equal instances
- [x] `bundle exec rspec spec/unit/stoplight/domain/failure_spec.rb`
- [x] Update `sig/_private/stoplight/domain/failure.rbs` with `eql?` / `hash`
- [x] Smoke in IRB: two equal Failures → `eql?` true, `hash` equal, `{a=>1,b=>2}.size == 1`

### DIFF
--- a/lib/stoplight/domain/failure.rb
+++ b/lib/stoplight/domain/failure.rb
@@ -39,6 +39,10 @@ module Stoplight
           error_message == other.error_message &&
           time == other.time
       end
+
+      alias_method :eql?, :==
+
+      def hash = [self.class, error_class, error_message, time].hash
     end
   end
 end
--- a/sig/_private/stoplight/domain/failure.rbs
+++ b/sig/_private/stoplight/domain/failure.rbs
@@ -11,6 +11,8 @@ module Stoplight
       def initialize: (String error_class, String error_message, Time time) -> void
 
       def ==: (untyped other) -> bool
+      def eql?: (untyped other) -> bool
+      def hash: () -> Integer
     end
   end
 end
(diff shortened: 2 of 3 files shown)


bolshakov (approved):

bolshakov:
Perfect!

PR #842, opened 2026-07-23
Rescue raising error_notifier in Notifier::FailSafe#notify

### LINKED ISSUE #775: `FailSafe#notify`'s fallback can itself raise and escape if `error_notifier` raises
## Summary

`Stoplight::Infrastructure::Notifier::FailSafe#notify`'s fallback proc calls `error_notifier.call(exception)` with no 
rescue around it. If a custom `error_notifier` itself raises (e.g. it does its own I/O or logs to a broken sink), that 
exception is not caught anywhere and escapes the wrapper -- defeating the documented "never propagates to the caller"
fail-safe invariant.

## Details

```ruby
fallback = proc do |exception|
  error_notifier.call(exception) if exception
  nil
end
circuit_breaker.run(fallback) do
  notifier.notify(info, from_color, to_color, error)
end
```

## Suggested fix

Wrap `error_notifier.call(exception)` in its own `begin/rescue StandardError` (swallow, or as a last resort
emit to `$stderr`) so a broken `error_notifier` can never escape the fallback. 

Do not introduce circuit breaker to guard error notifier, this is an overkill for its usage scenarios

johnsonsirv:
## Summary
- Wrap `error_notifier.call` in `Notifier::FailSafe#notify`'s fallback with `rescue` so a raising `error_notifier` cannot escape to the caller.
- Align with the existing telemetry pattern (`warn` when the error notifier itself raises).
- Fixes #775

## Test plan
- [x] Spec: real notifier raises and `error_notifier` raises → `#notify` does not raise
- [x] Spec: real notifier raises and `error_notifier` succeeds → still calls `error_notifier` (existing behavior)
- [x] Spec: successful notify unchanged
- [x] `bundle exec rspec spec/unit/stoplight/infrastructure/notifier/fail_safe_spec.rb`
- [x] Smoke: FailSafe with raising `error_notifier` returns without exception

### DIFF
--- a/lib/stoplight/infrastructure/notifier/fail_safe.rb
+++ b/lib/stoplight/infrastructure/notifier/fail_safe.rb
@@ -26,7 +26,13 @@ module Stoplight
         # Sends a notification using the wrapped notifier with fail-safe mechanisms.
         def notify(info, from_color, to_color, error = nil)
           fallback = proc do |exception|
-            error_notifier.call(exception) if exception
+            if exception
+              begin
+                error_notifier.call(exception)
+              rescue
+                # swallow
+              end
+            end
             nil
           end #: ^(StandardError?) -> void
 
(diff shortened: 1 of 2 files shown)


bolshakov:
Closed by https://github.com/bolshakov/stoplight/pull/843

PR #843, opened 2026-07-23
Rescue raising error_notifier in FailSafe storage wrappers

### LINKED ISSUE #773: Unguarded `error_notifier.call` defeats the fail-safe guarantee if the notifier itself raises
## Summary

Every fallback lambda in the five fail-safe storage wrappers calls `error_notifier.call(error)` without rescuing
it. Since `error_notifier` is user-supplied and can perform I/O (e.g. Sentry/Datadog/HTTP), a notifier failure
propagates straight out of the wrapper and into the caller, exactly defeating the component's purpose.

## Location

- `lib/stoplight/infrastructure/fail_safe/storage/*.rb`

## Details

The circuit breaker's own storage or notifier failure must never propagate to the caller. Today only the 
storage-failure half of that contract is honored -- if `error_notifier.call` itself raises, the new 
exception propagates directly out of `Metrics#record_failure`, `State#set_state`, etc., and into the caller's code.

`error_notifier` defaults to `->(error) { warn error }` but is freely overridable to something that performs
network I/O. This is precisely the correlated failure mode the component exists to protect against: a Redis
outage is exactly when a network-backed error notifier is also likely to fail.

The gap is unexercised by tests -- existing specs stub `error_notifier` with `receive(:call)` but never
`and_raise`.

## Suggested fix

Guard every `error_notifier.call(error)` invocation with a rescue (e.g. centralize it in the shared private
`fallback` helper) so a broken or I/O-performing notifier degrades to "unobserved" rather than crashing the
caller. 

Do not introduce circuit breaker to guard error notifier, this is an overkill for its usage scenarios

johnsonsirv:
## Summary
- Add `Infrastructure::FailSafe::ErrorNotifier` so a raising user `error_notifier` is swallowed and cannot escape fail-safe paths.
- Inject it once in `LightFactory` (same `#call` interface); storage and notifier FailSafe call sites stay unchanged.
- Introduce `Domain::_ErrorNotifier` (with Object methods like other ports) and point `Domain::error_notifier` at it so the wrapper can be injected like `Notifier::FailSafe`.
- Align `Telemetry::Bus` RBS ivar with the shared `error_notifier` type (types only; Bus still receives the raw config notifier).
- Left `GlobalState` / `FailSafe::Storage::Registry` alone for now to avoid conflicting with systems work.
- Fixes #773 (also covers the notifier FailSafe case from #775 for lights built via `LightFactory`).

## Test plan
- [x] `bundle exec rbs -I sig validate` (locally)
- [x] `bundle exec rspec` (full suite)
- [x] `STOPLIGHT_DATA_STORE=Memory bundle exec cucumber features/stoplight`
- [x] `STOPLIGHT_DATA_STORE=Redis bundle exec cucumber features/stoplight`

### DIFF
--- a/lib/stoplight/infrastructure/fail_safe/error_notifier.rb
+++ b/lib/stoplight/infrastructure/fail_safe/error_notifier.rb
@@ -1,0 +1,25 @@
+# frozen_string_literal: true
+
+module Stoplight
+  module Infrastructure
+    module FailSafe
+      # Wraps a user-provided error notifier so a failure in the notifier itself
+      # cannot escape to callers of fail-safe components.
+      #
+      # @api private
+      class ErrorNotifier
+        attr_reader :error_notifier
+
+        def initialize(error_notifier:)
+          @error_notifier = error_notifier
+        end
+
+        def call(error)
+          error_notifier.call(error)
+        rescue => notifier_error
+          warn "Stoplight error_notifier raised: #{notifier_error.message}"
+        end
+      end
+    end
+  end
+end
--- a/lib/stoplight/wiring/light_factory.rb
+++ b/lib/stoplight/wiring/light_factory.rb
@@ -17,7 +17,9 @@ module Stoplight
         @cool_off_time = config.cool_off_time
 
         @data_store_config = config.data_store
-        @error_notifier = config.error_notifier
+        @error_notifier = Infrastructure::FailSafe::ErrorNotifier.new(
+          error_notifier: config.error_notifier
+        )
         @notifiers = config.notifiers
         @traffic_recovery = config.traffic_recovery
         @traffic_control = config.traffic_control
--- a/sig/stoplight/domain/ports/error_notifier.rbs
+++ b/sig/stoplight/domain/ports/error_notifier.rbs
@@ -1,0 +1,12 @@
+module Stoplight
+  module Domain
+    # Called when Stoplight internals fail (e.g. Redis unavailable, notifier I/O).
+    interface _ErrorNotifier
+      def call: (StandardError) -> void
+
+      # Object methods
+      def ==: (untyped) -> bool
+      def is_a?: (untyped) -> bool
+    end
+  end
+end
--- a/sig/stoplight/domain/telemetry/bus.rbs
+++ b/sig/stoplight/domain/telemetry/bus.rbs
@@ -24,7 +24,7 @@ module Stoplight
         # The producer side (publish/subscribed?), what wiring injects into Emitter.
         include _TelemetryPublisher
 
-        @error_notifier: ^(StandardError) -> void
+        @error_notifier: error_notifier
         @mutex: Mutex
         @max_subscriptions: Integer
         @subscriptions: Array[[Subscription, subscription_filter, subscription_handler]]
--- a/sig/stoplight/domain/types.rbs
+++ b/sig/stoplight/domain/types.rbs
@@ -17,7 +17,7 @@ module Stoplight
     type state = unlocked | locked_green | locked_red
 
     type recovery_decision = :green | :yellow | :red
-    type error_notifier = ^(StandardError) -> void
+    type error_notifier = _ErrorNotifier
 
     # Percentage as decimal (0.0 to 1.0)
     type percentage = Float
--- a/sig/stoplight/infrastructure/fail_safe/error_notifier.rbs
+++ b/sig/stoplight/infrastructure/fail_safe/error_notifier.rbs
@@ -1,0 +1,13 @@
+module Stoplight
+  module Infrastructure
+    module FailSafe
+      class ErrorNotifier
+        include Domain::_ErrorNotifier
+
+        attr_reader error_notifier: Domain::error_notifier
+
+        def initialize: (error_notifier: Domain::error_notifier) -> void
+      end
+    end
+  end
+end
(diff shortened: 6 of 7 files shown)


bolshakov:
Hey Victor 👋 I see you made a few very similar changes and thinking what if we wrap user provided error notifier into a fail safe wrapper as we do for other expetrnal dependencies?

I hat would this do? We'll build dependency with exactly the same interface and inject it during the build time. This dependency will be responsible for calling error notifier and handling its error if any. Since the dependency is injected at light instantiation time, we won't need any changes on call sites a all!

What ro you think?

johnsonsirv:
Thanks @bolshakov , that makes sense, and it’s cleaner than repeating rescue at every call site (today & possibly in the future).

If I understand correctly: wrap the user-provided error_notifier once at build time in a small fail-safe collaborator with the same #call interface, inject that everywhere we currently pass error_notifier, and keep call sites as plain error_notifier.call(error). The wrapper would rescue/swallow (no circuit breaker, per the issue notes).

Happy to rework this PR that way (or open a follow-up that centralizes wrapping and reverts the per-site rescues). 

Which do you think ?

bolshakov:
Yes sounds like a plan!

bolshakov:
Hey @johnsonsirv how is it going? If you need any help, don't hesitate to ask - we'll be happy to assist

johnsonsirv:
> Hey @johnsonsirv how is it going? If you need any help, don't hesitate to ask - we'll be happy to assist

HI @bolshakov, thanks - i went on a short summer vacation 😄 


I would need your input on key decisions
Like we already agreed, wrapping the user error_notifier once (same #call interface) and injecting at build time so call sites stay unchanged,

There are 2 places to put that wrap, and I want your call before I rework (branch also has conflicts with develop).

The main value of Path A is scope control.

Path A: Storage construction only (#773 scope) - potentially wasted work
- Wrap once where storage FailSafe wrappers are built: Wiring::Redis::Backend (State / Metrics / RecoveryLock)
- GlobalState#create_registry (Registry)
- Touches: new FailSafe::ErrorNotifier + those 2 wiring sites.  
- Storage / notifier / telemetry source stay as-is.
- Does not fix: Notifier::FailSafe (#775), telemetry.
- Pros: ticket-scoped. Cons: second PR later to lift the wrap higher.


Path B: One wrap at system build (your vision)
- Wrap once in GlobalState#create_system via config.with(error_notifier: …) before registry + System / LightFactory.
- Touches: new wrapper + global_state.rb only (call sites unchanged). 
- Also covers: storage (#773) + notifier FailSafe (#775) + emitter/telemetry consumers of that config.
- Pros: one injection point, no call-site churn, closes both FailSafe tickets.
- Cons: slightly broader than #773, telemetry may stop warning when the notifier raises (wrapper swallows first). possibly warn in the wrapper like telemetry does today


I’d take Path B if you’re fine with that scope, then I can close the separate notifier FailSafe PR as superseded.

Happy to rework once i get your feedback

bolshakov:
Hey, hope you had a great vacation! 

First of all, regarding the conflict with develop branch - this should not be a big deal as i just moved these files (removed `_private` folder):
- `sig/_private/stoplight/infrastructure/fail_safe/storage/recovery_lock.rbs` -> `sig/stoplight/infrastructure/fail_safe/storage/recovery_lock.rbs`
- `sig/_private/stoplight/infrastructure/fail_safe/storage/registry.rbs` -> `sig/stoplight/infrastructure/fail_safe/storage/registry.rbs`

I also lean toward B, but let's do a small simplification -let's not touch `Stoplight::Wiring::GlobalState` for now and do this wrapping in the `Stoplight::Wiring::LightFactory`. It's not like systems and GlobalState don't need this, it's because I change things around systems and don't want another conflict

johnsonsirv:
> Hey, hope you had a great vacation!
> 
> First of all, regarding the conflict with develop branch - this should not be a big deal as i just moved these files (removed `_private` folder):
> 
> * `sig/_private/stoplight/infrastructure/fail_safe/storage/recovery_lock.rbs` -> `sig/stoplight/infrastructure/fail_safe/storage/recovery_lock.rbs`
> * `sig/_private/stoplight/infrastructure/fail_safe/storage/registry.rbs` -> `sig/stoplight/infrastructure/fail_safe/storage/registry.rbs`
> 
> I also lean toward B, but let's do a small simplification -let's not touch `Stoplight::Wiring::GlobalState` for now and do this wrapping in the `Stoplight::Wiring::LightFactory`. It's not like systems and GlobalState don't need this, it's because I change things around systems and don't want another conflict


Sounds good! Thanks for your feedback

johnsonsirv:
Hi 👋  @bolshakov Kindly review 👀 

Reworked from per-call-site rescues to a single injected wrapper . I’ll close the separate notifier FailSafe PR as superseded by this.

lib/stoplight/infrastructure/fail_safe/error_notifier.rb
```diff
@@ -0,0 +1,25 @@
...
+        attr_reader :error_notifier
+
+        def initialize(error_notifier:)
+          @error_notifier = error_notifier
+        end
+
+        def call(error)
+          error_notifier.call(error)
+        rescue
+          # swallow
```
bolshakov: please call `warn` here because failed error notifier level no trace at all

bolshakov (commented):
Great job! I left a small request

lib/stoplight/infrastructure/fail_safe/error_notifier.rb
johnsonsirv: Resolved in https://github.com/bolshakov/stoplight/commit/b7614d466cf6b60ffeda585260d8fb71dcbe63c9

bolshakov (approved):
Thank you for your contribution! It was quite a lot of back-and-forth here but I'm really happy with the final result!

PR #846, opened 2026-07-26
fix: Prevent unbounded retries in Scripting#call on persistent NOSCRIPT errors

### LINKED ISSUE #777: Unbounded, backoff-free retry loop in `Scripting#call` on persistent `NOSCRIPT`
## Summary
`Scripting#call` retries indefinitely on any `NOSCRIPT` `Redis::CommandError`, with no attempt counter and no
backoff. If `NOSCRIPT` keeps recurring, the calling thread spins against Redis forever instead of failing fast.

## Location
`lib/stoplight/infrastructure/redis/storage/scripting.rb:21-32`

## Details
```ruby
rescue ::Redis::CommandError => error
  if error.message.include?("NOSCRIPT")
    reload_script(script_name)
    retry
  else
    raise
  end
end
```
In the ordinary case (one node evicts the script) a single retry after `SCRIPT LOAD` fixes it, and that path is
tested (`scripting_spec.rb:38-46`). But if `NOSCRIPT` keeps recurring -- e.g. Redis Cluster resharding/failover
landing requests on a node that never got the script loaded, or a shared instance under repeated
`SCRIPT FLUSH` -- this becomes a tight, immediate retry loop with no cap and no backoff. Because the method
never raises in that scenario, the outer `FailSafe` circuit breaker never observes a failure and can't degrade
to the memory fallback -- the calling thread just spins against Redis instead of failing fast into the
documented graceful-degradation path.

## Suggested fix

Cap retries (e.g. 3 attempts) and raise after exhausting them so `FailSafe` can catch it and fail over; add a
short backoff if more than one retry is ever justified.

arvindpz:
Fixes https://github.com/bolshakov/stoplight/issues/777

### Problem description:
`Scripting#call` could retry indefinitely on `NOSCRIPT` errors as long as scripts weren't loaded. It did not gracefully fall back to the Failsafe store as documented.

### Changes made:
- Upon persistent `NOSCRIPT` errors from Redis, do not retry more than 1 (`REDIS_NOSCRIPT_MAX_RETRY`) times. The retry doesn't include the first call made to Redis
- ~The retries have delays of 0.2, 0.4, and 0.8 seconds between each attempt -- a total of 1.4 seconds before falling back to failsafe.~ Backoff was not helpful and hence decided unnecessary. See https://github.com/bolshakov/stoplight/pull/846#discussion_r3654726364
- Added specs covering both scenarios -- recovery within allowed retries and exhaustion after allowed retries

### Other changes:
- Fixed a minor typo in wiring/system.rb (not relevant to the issue, but still included it)

### DIFF
--- a/lib/stoplight/infrastructure/redis/storage/scripting.rb
+++ b/lib/stoplight/infrastructure/redis/storage/scripting.rb
@@ -11,6 +11,9 @@ module Stoplight
           INCLUDE_DIRECTIVE = /^--\s*@include\s+([\w\-\/]+)$/
           private_constant :INCLUDE_DIRECTIVE
 
+          REDIS_NOSCRIPT_MAX_RETRY = 1
+          private_constant :REDIS_NOSCRIPT_MAX_RETRY
+
           class << self
             def default_scripts_root = SCRIPTS_ROOT
           end
@@ -22,15 +25,21 @@ module Stoplight
           end
 
           def call(script_name, keys: [], args: [])
-            redis.then do |client|
-              client.evalsha(script_sha(script_name), keys: keys.map(&:to_s), argv: args.map(&:to_s))
-            end
-          rescue ::Redis::CommandError => error
-            if error.message.include?("NOSCRIPT")
-              reload_script(script_name)
-              retry
-            else
-              raise
+            retries = 0
+
+            begin
+              redis.then do |client|
+                client.evalsha(script_sha(script_name), keys: keys.map(&:to_s), argv: args.map(&:to_s))
+              end
+            rescue ::Redis::CommandError => error
+              if error.message.include?("NOSCRIPT") && retries < REDIS_NOSCRIPT_MAX_RETRY
+                retries += 1
+                warn "Stoplight is unable to find the script '#{script_name}' in Redis's script cache. Reloading and retrying..."
+                reload_script(script_name)
+                retry
+              else
+                raise error
+              end
             end
           end
 
--- a/lib/stoplight/infrastructure/redis/storage/scripting.rb
+++ b/lib/stoplight/infrastructure/redis/storage/scripting.rb
@@ -11,9 +11,6 @@ module Stoplight
           INCLUDE_DIRECTIVE = /^--\s*@include\s+([\w\-\/]+)$/
           private_constant :INCLUDE_DIRECTIVE
 
-          REDIS_NOSCRIPT_MAX_RETRY = 1
-          private_constant :REDIS_NOSCRIPT_MAX_RETRY
-
           class << self
             def default_scripts_root = SCRIPTS_ROOT
           end
@@ -25,15 +22,15 @@ module Stoplight
           end
 
           def call(script_name, keys: [], args: [])
-            retries = 0
+            retried = false
 
             begin
               redis.then do |client|
                 client.evalsha(script_sha(script_name), keys: keys.map(&:to_s), argv: args.map(&:to_s))
               end
             rescue ::Redis::CommandError => error
-              if error.message.include?("NOSCRIPT") && retries < REDIS_NOSCRIPT_MAX_RETRY
-                retries += 1
+              if error.message.include?("NOSCRIPT") && !retried
+                retried = true
                 warn "Stoplight is unable to find the script '#{script_name}' in Redis's script cache. Reloading and retrying..."
                 reload_script(script_name)
                 retry
--- a/lib/stoplight/wiring/system.rb
+++ b/lib/stoplight/wiring/system.rb
@@ -51,7 +51,7 @@ module Stoplight
       #   of this system. For example if your system uses Redis data store, or notifiers that
       #   communicate with external systems, they could go off. Failover system hosts
       #   all the circuit breakers that are needed to protect these dependencies from failing.
-      #   Failover system itself never uses external dependencies and thereforec does not need
+      #   Failover system itself never uses external dependencies and therefore does not need
       #   external failover.
       def initialize(config:, failover_system:, registry:)
         @name = config.name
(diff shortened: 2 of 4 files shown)


lib/stoplight/infrastructure/redis/storage/scripting.rb
```diff
@@ -22,15 +28,22 @@ def initialize(redis:, scripts_root: self.class.default_scripts_root)
...
-              raise
+            retries = 0
+
+            begin
+              redis.then do |client|
+                client.evalsha(script_sha(script_name), keys: keys.map(&:to_s), argv: args.map(&:to_s))
+              end
+            rescue ::Redis::CommandError => error
+              if error.message.include?("NOSCRIPT") && retries < REDIS_NOSCRIPT_MAX_RETRY
+                delay = REDIS_NOSCRIPT_BACKOFF * (2**retries)
```
arvindpz: In the worst case, this may take ~1.4 seconds before falling back to FailSafe. Please let me know if you'd prefer shorter / flat delays.

spec/unit/stoplight/infrastructure/redis/storage/scripting_spec.rb
```diff
@@ -186,4 +186,53 @@
...
+
+    it "does not propagate error if retries succeed within REDIS_NOSCRIPT_MAX_RETRY attempts" do
+      responses = [
+        -> { raise ::Redis::CommandError.new("NOSCRIPT") },    # Initial call fails
+        -> { raise ::Redis::CommandError.new("NOSCRIPT") },    # First retry fails
+        -> { raise ::Redis::CommandError.new("NOSCRIPT") },    # Second retry fails
+        -> { "success" }                                       # Third retry succeeds
+      ]
+
+      allow(redis).to receive(:evalsha) { responses.shift.call }
```
bolshakov, line 200: This is clever!

lib/stoplight/wiring/system.rb
```diff
@@ -51,7 +51,7 @@ class System
       #   of this system. For example if your system uses Redis data store, or notifiers that
       #   communicate with external systems, they could go off. Failover system hosts
       #   all the circuit breakers that are needed to protect these dependencies from failing.
-      #   Failover system itself never uses external dependencies and thereforec does not need
+      #   Failover system itself never uses external dependencies and therefore does not need
```
bolshakov, line 54: Good catch

lib/stoplight/infrastructure/redis/storage/scripting.rb
bolshakov: The formula's a good start but I have to request a few changes. Re-reading my own ticket, I wasn't sure about backoff ("add one if more than one retry is ever justified") and walking through the scenarios that could happen, I don't think it ever is:

1. Ordinary case: single node evicts the script: reload, retry once, done. Deterministic, no wait needed and waiting first doesn't make the reload any more likely to succeed.
2. Cluster resharding/failover landing on a node that never got the script loaded. The script cache is per-node, so this either resolves on the very next attempt (once routed to a node that has it) or it's still misrouted. Real resharding/failover takes far longer than any short backoff we'd add here. A few hundred ms of waiting between retries doesn't give the cluster time to actually finish rebalancing; it just delays reaching the same outcome. If it's still wrong after one retry, we should fail into FailSafe's fallback, not keep spinning.
3. Shared instance under repeated `SCRIPT FLUSH`. If flushing is happening fast enough that even one reload+retry doesn't survive it, that's an external process aggressively flushing scripts. This is a systemic issue that not any number of retry can solve. More attempts (backed off or not) just hammer an already-misbehaving Redis a few extra times before hitting the same outcome - raising an error.

In all three scenarios, either the first retry succeeds or the situation is one where retrying further (with or without a delay) doesn't change the outcome, just delays it. So, my ask:

- Drop the delay entirely - it never buys us anything.
- Reduce `REDIS_NOSCRIPT_MAX_RETRY` to a single retry (reload once, retry once, raise if it still fails) - nothing in the scenarios above is fixed by a 2nd or 3rd attempt.
- As you're here could you log (use just `warn message`) on the retry attempt so it's not invisible in production

Sorry for the back-and-forth - the backoff was in my own ticket, but on reflection I don't think it's worth keeping it.

spec/unit/stoplight/infrastructure/redis/storage/scripting_spec.rb
arvindpz: Thank you!

lib/stoplight/infrastructure/redis/storage/scripting.rb
arvindpz: Thanks for the thorough review and for taking the time to elucidate the scenarios we are fixing. I wasn't very sure about the exact scenarios where this issue could've happened in production, and your comment gives me a clear picture.
Thanks for pushing me towards the right approach.

lib/stoplight/infrastructure/redis/storage/scripting.rb
```diff
@@ -22,15 +25,21 @@ def initialize(redis:, scripts_root: self.class.default_scripts_root)
...
-            else
-              raise
+            retries = 0
+
+            begin
+              redis.then do |client|
+                client.evalsha(script_sha(script_name), keys: keys.map(&:to_s), argv: args.map(&:to_s))
+              end
+            rescue ::Redis::CommandError => error
+              if error.message.include?("NOSCRIPT") && retries < REDIS_NOSCRIPT_MAX_RETRY
```
bolshakov: Thank you for addressing feedback!

I think after these changes the old structure tailored for multiple retry attempts is not needed here anymore and the same idea could be expressed with just a boolean flag

```
retried = false

begin
  load
Rescue
  if no script && !retried
    retried = true
    retry
  Else
    Raise
  End
End
```

Would it make more sense?

arvindpz: Makes total sense, thanks! I have updated the PR with this simplification.

bolshakov (approved):

bolshakov:
Great job 🚀

PR #852, opened 2026-07-28
 Wire RecoverySucceeded event (#744)

### LINKED ISSUE #744: Wire `RecoverySucceeded` event

`RecoverySucceeded`.

## What you're wiring up

`RecoverySucceeded` (`lib/stoplight/domain/telemetry/recovery_succeeded.rb`,
`sig/stoplight/telemetry/recovery_succeeded.rbs`) fires when the recovery policy decides the
light has proven itself and resumes normal traffic (yellow -> green).

## Where to make the change

`Domain::Tracker::RecoveryProbe#recover`, the `to_color == Color::GREEN` branch:

```ruby
def recover
  recovery_metrics = metrics_store.metrics_snapshot
  recovery_result = traffic_recovery.determine_color(config, recovery_metrics)

  return if recovery_result == TrafficRecovery::YELLOW

  from_color, to_color = case recovery_result
  when TrafficRecovery::GREEN then [Color::YELLOW, Color::GREEN]
  when TrafficRecovery::RED then [Color::YELLOW, Color::RED]
  else raise "..."
  end

  state_store.transition_to_color(to_color)   # <- return value currently discarded
  metrics_store.clear
  notifiers.each { |notifier| notifier.notify(info, from_color, to_color, nil) }
end
```

**Heads up, this is the trickiest part of this issue:** `state_store.transition_to_color`
returns `true` only the first time it actually flips the color -- but this method throws that
return value away and calls `notify` unconditionally.  This seems a bug (I'll create a ticket for it separately)

```ruby
transitioned = state_store.transition_to_color(to_color)
metrics_store.clear
notifiers.each { ... }   # left exactly as-is, still unconditional
# emit RecoverySucceeded only if transitioned
```

## Payload fields

- `from_color:` -- always `Color::YELLOW`.
- `to_color:` -- always `Color::GREEN`.
- `policy:` -- a string the traffic-recovery policy uses to self-identify (e.g.
  `"consecutive_successes"`). Check what `traffic_recovery` currently exposes for this.
- `metrics:` -- the `recovery_metrics` local already computed above.

## Suggested approach

1. `Tracker::RecoveryProbe` needs the emitter constructor-injected -- the same way
   `Tracker::Request` will for `TrafficBreached`, if that ticket landed first, follow its
   wiring shape.
2. Add spec coverage in `spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb`: a probe
   that flips yellow -> green emits exactly one `RecoverySucceeded`; a probe that keeps the
   light yellow emits none.

## Out of scope

- Fixing the pre-existing unconditional-notify asymmetry vs. `Tracker::Request` -- flag it if you
  like, but leave the behavior alone.
- `RecoveryFailed` (the `to_color == RED` branch of this same method) -- separate ticket.

## References

- `sig/stoplight/telemetry/recovery_succeeded.rbs`

Zainulhassan01:
Fixes #744, Wires up the `RecoverySucceeded` telemetry event so it fires when a light recovers

What I did:
1. `RecoveryProbe` now takes an `emitter`, and emits `RecoverySucceeded` when it successfully transitions the light to green.
2. The emit sits *inside* the `state_store.transition_to_color(...)` guard, so it only fires when this probe actually won the transition.
3.  Added the matching RBS for the new `emitter` param/reader.

**Policy name is hardcoded** to `"consecutive_successes"`. Right now that's the only recovery policy that exists, so I didn't want to build indirection for a single case. Happy to wire it to the actual policy if/when there's more than one.

### DIFF
--- a/lib/stoplight/domain/tracker/recovery_probe.rb
+++ b/lib/stoplight/domain/tracker/recovery_probe.rb
@@ -4,12 +4,13 @@ module Stoplight
   module Domain
     module Tracker
       class RecoveryProbe
-        def initialize(traffic_recovery:, notifiers:, config:, metrics_store:, state_store:)
+        def initialize(traffic_recovery:, notifiers:, config:, metrics_store:, state_store:, emitter:)
           @traffic_recovery = traffic_recovery
           @notifiers = notifiers
           @config = config
           @metrics_store = metrics_store
           @state_store = state_store
+          @emitter = emitter
         end
 
         # @param exception [Exception]
@@ -32,6 +33,7 @@ module Stoplight
         attr_reader :config
         attr_reader :metrics_store
         attr_reader :state_store
+        attr_reader :emitter
 
         def recover
           recovery_metrics = metrics_store.metrics_snapshot
@@ -51,6 +53,18 @@ module Stoplight
             info = LightInfo.new(name: config.name)
             notifiers.each do |notifier|
               notifier.notify(info, from_color, to_color, nil)
+            end
+            if to_color.to_sym == TrafficRecovery::GREEN
+              emitter.emit(Telemetry::RecoverySucceeded) do
+                Telemetry::RecoverySucceeded.new(from_color: Color::YELLOW, to_color: Color::GREEN,
+                  policy: TrafficRecovery::ConsecutiveSuccesses::NAME.to_s,
+                  metrics: Telemetry::Metrics.new(
+                    successes: recovery_metrics.successes,
+                    errors: recovery_metrics.errors,
+                    consecutive_errors: recovery_metrics.consecutive_errors,
+                    consecutive_successes: recovery_metrics.consecutive_successes
+                  ))
+              end
             end
           end
         end
--- a/lib/stoplight/domain/tracker/recovery_probe.rb
+++ b/lib/stoplight/domain/tracker/recovery_probe.rb
@@ -41,23 +41,13 @@ module Stoplight
 
           return if recovery_result == TrafficRecovery::YELLOW
 
-          from_color, to_color = case recovery_result
-          when TrafficRecovery::GREEN then [Color::YELLOW, Color::GREEN]
-          when TrafficRecovery::RED then [Color::YELLOW, Color::RED]
-          else
-            raise "recovery strategy returned unexpected color: #{recovery_result}"
-          end
-
-          if state_store.transition_to_color(to_color)
-            metrics_store.clear
-            info = LightInfo.new(name: config.name)
-            notifiers.each do |notifier|
-              notifier.notify(info, from_color, to_color, nil)
-            end
-            if to_color.to_sym == TrafficRecovery::GREEN
+          case recovery_result
+          when TrafficRecovery::GREEN
+            to_color = Color::GREEN
+            if transition(from_color: Color::YELLOW, to_color:)
               emitter.emit(Telemetry::RecoverySucceeded) do
-                Telemetry::RecoverySucceeded.new(from_color: Color::YELLOW, to_color: Color::GREEN,
-                  policy: TrafficRecovery::ConsecutiveSuccesses::NAME.to_s,
+                Telemetry::RecoverySucceeded.new(from_color: Color::YELLOW, to_color:,
+                  policy: traffic_recovery.name,
                   metrics: Telemetry::Metrics.new(
                     successes: recovery_metrics.successes,
                     errors: recovery_metrics.errors,
@@ -66,7 +56,21 @@ module Stoplight
                   ))
               end
             end
+          when TrafficRecovery::RED
+            transition(from_color: Color::YELLOW, to_color: Color::RED)
+          else
+            raise "recovery strategy returned unexpected color: #{recovery_result}"
           end
+        end
+
+        def transition(from_color:, to_color:)
+          return false unless state_store.transition_to_color(to_color)
+          metrics_store.clear
+          info = LightInfo.new(name: config.name)
+          notifiers.each do |notifier|
+            notifier.notify(info, from_color, to_color, nil)
+          end
+          true
         end
       end
     end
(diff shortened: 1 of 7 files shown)


bolshakov:
Please rebase on top of the `develop` branch. What I did there: moved `sig/_private/stoplight/domain/tracker/recovery_probe.rbs` to `sig/stoplight/domain/tracker/recovery_probe.rbs` so the fix should be mechanical

lib/stoplight/domain/tracker/recovery_probe.rb
```diff
@@ -52,6 +54,18 @@ def recover
             notifiers.each do |notifier|
               notifier.notify(info, from_color, to_color, nil)
             end
+            if to_color.to_sym == TrafficRecovery::GREEN
+              emitter.emit(Telemetry::RecoverySucceeded) do
+                Telemetry::RecoverySucceeded.new(from_color: Color::YELLOW, to_color: Color::GREEN,
+                  policy: TrafficRecovery::ConsecutiveSuccesses::NAME.to_s,
```
bolshakov: This hardcodes the policy name to `ConsecutiveSuccesses`. We only have one recovery strategy today, but we're planning to add more ("error rate" - the one I have in mind - recevers then error rate falls below specific threshold), and this code should be ready for that rather than needing a rewrite later. 

One option: add a `#name` (or similar) method to the `_TrafficRecovery` port, so each strategy can report its own name, and this line becomes `traffic_recovery.name` instead of a hardcoded reference.

lib/stoplight/domain/tracker/recovery_probe.rb
```diff
@@ -52,6 +54,18 @@ def recover
             notifiers.each do |notifier|
               notifier.notify(info, from_color, to_color, nil)
             end
+            if to_color.to_sym == TrafficRecovery::GREEN
```
bolshakov: Small simplification: `recovery_result` (from the case a few lines up) is already `TrafficRecovery::GREEN` or `TrafficRecovery::RED`  right here, so you can write `if recovery_result == TrafficRecovery::GREEN` instead of converting `to_color` back into a symbol to compare it. 

`Color` and `TrafficRecovery` are separate vocabularies on purpose -- Color is the light's actual state, while `TrafficRecovery` is the recovery strategy's decision about what to do next:
- green - recovered successful 
- yellow -  keep probing, no enough information to make decision yet
- red - recovery failed

I know they it's a bit confusing and probably we should use better naming in the future.

lib/stoplight/domain/tracker/recovery_probe.rb
```diff
@@ -52,6 +54,18 @@ def recover
...
+              emitter.emit(Telemetry::RecoverySucceeded) do
+                Telemetry::RecoverySucceeded.new(from_color: Color::YELLOW, to_color: Color::GREEN,
```
bolshakov:
```suggestion
                Telemetry::RecoverySucceeded.new(from_color:, to_color:),
```

let's just reuse already set variables

lib/stoplight/domain/tracker/recovery_probe.rb
bolshakov: Create a separate ticket for this https://github.com/bolshakov/stoplight/issues/853

spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb
```diff
@@ -107,4 +115,69 @@
 
     include_examples "recovering after probe"
   end
+
+  describe "RecoverySucceeded telemetry" do
```
bolshakov, line 121: This block covers "stays yellow" and "fails to transition to green" as negative cases, but there's no test asserting that recovering to `RED` doesn't emit `RecoverySucceeded`. It's exercised by the existing "when recover to RED" shared example elsewhere in this file, but that one doesn't check emitter.emitted, so a regression that fired this event on the RED path too would slip through. 

Could you add a context "when the probe changes from yellow to red" here, mirroring the other two `not_to emit(...)` cases?

bolshakov (commented):
Great job overall! I left a few comments

lib/stoplight/domain/tracker/recovery_probe.rb
Zainulhassan01: I remember I used the variables here but getting steep error using `to_color` because steep can't tell whether it's green or red(the above case leaves it a "green" | "red" union), unlike `from_color`(it narrows to yellow from the case above). So I used both values directly, should I just reuse `from_color` variable? or do you have any other option in mind?

bolshakov: Narrow types bit us here! What if we move the whole branch under corresponding when, keeping each "when" fgully self-contained:

```ruby
case recovery_result
when TrafficRecovery::GREEN
  to_color = Color::GREEN
  ... state_store.transition_to_color(to_color) ...
  Telemetry::RecoverySucceeded.new(from_color: Color::YELLOW, to_color:, policy:, metrics:)
when TrafficRecovery::RED
  to_color = Color::RED
  ...
  Telemetry::RecoveryFailed.new(from_color: Color::YELLOW, to_color:, policy:, failure:, metrics:)
end
```

this basically uses the shape you originally implemented here, but moves it up into branch where types could be more specfic and there is less room for future errors as branching happens only i none place. 

The integration of `RecoveryFailed` event will be trivial - this is a separate ticket though

lib/stoplight/domain/tracker/recovery_probe.rb
```diff
@@ -32,27 +33,44 @@ def record_success
...
           end
+        end
 
-          if state_store.transition_to_color(to_color)
-            metrics_store.clear
-            info = LightInfo.new(name: config.name)
-            notifiers.each do |notifier|
-              notifier.notify(info, from_color, to_color, nil)
-            end
+        def transition(from_color:, to_color:)
```
bolshakov, line 66: Nice!

bolshakov (approved):

bolshakov (approved):
Amazing job, thank you 🚀

PR #865, opened 2026-08-01
Feature/admin ids instead of names

bolshakov:


bolshakov:
Renamed branch https://github.com/bolshakov/stoplight/pull/866

PR #870, opened 2026-08-07
chore: Make meaningful default exception message

Lokideos:
# Description

This PR aims to make meaningful deafult exception message for the Stoplight RedLight Exception.
At the moment, we return light name in the exception message, which might confuse people who rescue exception and send them to Sentry in a way like this:
```ruby
def call
# ...
rescue Stoplight::Error::RedLight => e
  IssueReporter.warn("Failure", message: e.message)
end
```
Given that Stoplight provides sensible defaults, I believe that we should provide a more meaningful default error message.

I suggest using something like `Stoplight "%{light_name}" is red - network traffic stopped until recovery.`.

### DIFF
--- a/lib/stoplight/error.rb
+++ b/lib/stoplight/error.rb
@@ -18,7 +18,7 @@ module Stoplight
     end
 
     class RedLight < Base
-      EXCEPTION_MESSAGE = 'Stoplight "%{light_name}" is red - network traffic stopped until recovery.'
+      EXCEPTION_MESSAGE = 'Stoplight "%{light_name}" is red - traffic stopped until recovery.'
       private_constant :EXCEPTION_MESSAGE
 
       # @!attribute light_name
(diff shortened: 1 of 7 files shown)


lib/stoplight/error.rb
```diff
@@ -18,6 +18,9 @@ class TooManySubscriptions < Base
     end
 
     class RedLight < Base
+      EXCEPTION_MESSAGE = 'Stoplight "%{light_name}" is red - network traffic stopped until recovery.'
```
bolshakov:
```suggestion
      EXCEPTION_MESSAGE = 'Stoplight "%{light_name}" is red - traffic stopped until recovery.'
```
Lets loose "network" - it's not always about network and in some contexts could he confusing

Lokideos: Totally agree, let's correct the message.

bolshakov (approved):

bolshakov:
Amazing, really good improvement

PR #873, opened 2026-08-07
Age the Memory metrics window on the monotonic clock

### LINKED ISSUE #779: WindowMetrics#record_success/#record_failure call the clock twice per invocation, doubling Time allocation on the hot path
## Summary
`WindowMetrics#record_success` and `#record_failure` each call `clock.current_time` twice per invocation instead
of once, doubling `Time` allocation on the metrics-recording path exercised by every rate-based (`error_rate`)
circuit-breaker strategy on every protected call, success or failure.

## Location
`lib/stoplight/infrastructure/memory/storage/window_metrics.rb:34-47` (`record_success`),
`lib/stoplight/infrastructure/memory/storage/window_metrics.rb:50-66` (`record_failure`)

## Details

`record_success` calls `@successes.increment` (line 36), which independently calls `clock.current_time` inside
`SlidingWindow#current_bucket` (`window_metrics/sliding_window.rb:64-70`), then fetches `current_time =
@clock.current_time` again (line 38) purely for the later `last_success_at` comparison. `record_failure` has the
same shape: `@errors.increment` triggers one `clock.current_time` call, and `Domain::Failure.from_error(exception,
time: @clock.current_time)` triggers a second, independent one.

Contrast with `UnboundedMetrics#record_success`/`#record_failure` (`unbounded_metrics.rb:34-46,48-65`), which fetch
the clock once and reuse the value.

Measured (benchmark-ips, report author's machine):
- `Time.now.utc` alone (what `SystemClock#current_time` does): ~108 ns/call.
- `WindowMetrics#record_success` (2 clock calls): ~438-460 ns/call.
- `UnboundedMetrics#record_success` (1 clock call): ~261 ns/call.
- A patched variant of `WindowMetrics` that fetches the clock once and threads the value into
  `SlidingWindow#increment(time)` measured 1.50x faster (438 ns -> 292 ns) for `record_success`.

This is the metrics-recording path exercised by every rate-based (`error_rate`) circuit-breaker strategy on
every protected call, success or failure, so it is genuinely hot, not speculative.

## Suggested fix

Compute `current_time` once per call and thread it through -- change `SlidingWindow#increment` to accept a
`time` parameter (using `time.to_i` for the bucket key, matching how `sum_in_window` already takes a
`window_start` parameter), and pass the same value into `Domain::Failure.from_error(exception, time:
current_time)`.

OursCodeur:
Closes #779

`SlidingWindow` now owns the clock domain: buckets key on monotonic seconds and `sum_in_window(window_size)` takes a duration, so callers cannot feed wall timestamps into the window. `WindowMetrics#build_metrics_snapshot` loses its wall read; `record_success` and `record_failure` are unchanged from develop.

Reason : Wall clock can corrupt FIFO eviction: a backward NTP step inserts an older-keyed bucket behind newer ones where `slide_window!` never reaches it, so it outlives the window. Monotonic keys make the nondecreasing invariant hold by construction.

Re: the original goal, each record op reads the wall clock exactly once, now pinned by specs. The internal monotonic reads are allocation-free and cheap (144ns vs 345ns for `Time.now.utc`).

Benchs:

| op | time | allocs/op |
|---|---|---|
| record_success | 1.2µs -> 1.1µs | 2 -> 1 |
| record_failure | 4.5–4.8µs -> 4.3µs | 15.1 -> 12.1 |
| metrics_snapshot | 2.7µs -> 2.5µs | 10 -> 8 |
| Light#run success path | parity | 5.0 -> 4.0 |

Tests:

A regression spec pins that events expire by elapsed time when the wall clock steps backwards. The suites set `Timecop.mock_process_clock = true` so travel and freeze move the monotonic lane; the "outside the running window" shared examples use timecop's relative form (`Timecop.freeze(-window_size - 10)`) because freezing to a `Time` target deliberately leaves the monotonic clock unshifted. 

`system_clock_spec`'s monotonic example no longer freezes: with the process clock mocked, a frozen assertion compares the mock to itself. `sliding_window_spec` drives the injected clock to absolute seconds instead of nesting Timecop blocks.

### DIFF
--- a/lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb
+++ b/lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb
@@ -26,14 +26,14 @@ module Stoplight
               @clock = clock
             end
 
-            # Increment the count at a given timestamp
+            # Increment the count at the current monotonic second
             def increment
               @buckets[current_bucket] += 1
               @running_sum += 1
             end
 
-            def sum_in_window(window_start)
-              slide_window!(window_start)
+            def sum_in_window(window_size)
+              slide_window!(monotonic_seconds - window_size)
               @running_sum
             end
 
@@ -58,11 +58,13 @@ module Stoplight
             end
 
             def current_bucket
-              bucket_for_time(@clock.current_time)
+              monotonic_seconds.to_i
             end
 
-            def bucket_for_time(time)
-              time.to_i
+            # Monotonic, so a wall-clock step (NTP) cannot break FIFO eviction;
+            # the clock port returns float milliseconds.
+            def monotonic_seconds
+              @clock.monotonic_time / 1000.0
             end
           end
         end
--- a/spec/unit/stoplight/infrastructure/memory/storage/window_metrics/sliding_window_spec.rb
+++ b/spec/unit/stoplight/infrastructure/memory/storage/window_metrics/sliding_window_spec.rb
@@ -1,69 +1,67 @@
 # frozen_string_literal: true
 
 RSpec.describe Stoplight::Infrastructure::Memory::Storage::WindowMetrics::SlidingWindow do
-  let(:counter) { described_class.new(clock:) }
-  let(:clock) { Stoplight::Infrastructure::SystemClock.new }
+  subject(:counter) { described_class.new(clock:) }
 
-  around do |example|
-    Timecop.freeze do
-      example.run
-    end
+  let(:clock) { instance_double(NullClock) }
+
+  # Position the monotonic clock at an absolute second.
+  def at(seconds)
+    allow(clock).to receive(:monotonic_time).and_return(seconds * 1000.0)
   end
 
   describe "#increment" do
     it "increments the count for the given time" do
+      at(100)
       counter.increment
       counter.increment
       counter.increment
 
-      expect(counter.sum_in_window(Time.now - 10)).to eq(3)
+      expect(counter.sum_in_window(10)).to eq(3)
     end
 
     it "when empty returns zero sum" do
-      expect(counter.sum_in_window(Time.now)).to eq(0)
-      expect(counter.sum_in_window(Time.now - 100)).to eq(0)
-      expect(counter.sum_in_window(Time.now - 10000)).to eq(0)
+      at(100)
+      expect(counter.sum_in_window(0)).to eq(0)
+      expect(counter.sum_in_window(100)).to eq(0)
+      expect(counter.sum_in_window(10000)).to eq(0)
     end
 
     it "increments the count for the different seconds" do
-      Timecop.freeze(Time.now - 2) do
-        counter.increment
-        counter.increment
-      end
-      Timecop.freeze(Time.now - 1) do
-        counter.increment
-        counter.increment
-      end
-      Timecop.freeze(Time.now) do
-        counter.increment
-      end
+      at(98)
+      counter.increment
+      counter.increment
+      at(99)
+      counter.increment
+      counter.increment
+      at(100)
+      counter.increment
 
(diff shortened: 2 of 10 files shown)


spec/unit/stoplight/infrastructure/memory/storage/window_metrics/sliding_window_spec.rb
```diff
@@ -1,69 +1,51 @@
...
-      expect(counter.sum_in_window(Time.now)).to eq(1)
+      counter.increment(current_time - 2)
+      counter.increment(current_time - 2)
+      counter.increment(current_time - 1)
+      counter.increment(current_time - 1)
+      counter.increment(current_time)
+
+      expect(counter.sum_in_window(current_time - 2)).to eq(5)
+      expect(counter.sum_in_window(current_time - 1)).to eq(3)
+      expect(counter.sum_in_window(current_time)).to eq(1)
```
bolshakov: absolutely love this change

lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb
```diff
@@ -18,17 +18,18 @@ class WindowMetrics
...
               # The running sum of all increments in the current window
               @running_sum = 0
-              @clock = clock
             end
 
             # Increment the count at a given timestamp
-            def increment
-              @buckets[current_bucket] += 1
+            #
+            # @note Times must be nondecreasing; FIFO eviction assumes chronological inserts.
```
bolshakov: Sharp observation. We use UTC time which never moves backwards, yet NTP adjustments could cause time move in any direction. Normally this happens by slowing/speeding up time, but bigger leaps are possible too. 

This is harmless, when happens forward - e.g. when you open a laptop after hibernation but breaks the window if it happens in an opposite direction, until window expire. 

This started as an effort to reduce allocations, but you pointed out here that maybe we should not use wall clock time here at all and the better feat would be to use monotonic seconds instead

 @OursCodeur how do you feel about this change?

OursCodeur: Agreed, eviction should run on elapsed time, backwards step appends an older-keyed bucket behind newer ones, `slide_window!` never reaches it, and it outlives the window.

I'd use monotonic_time for insertion and eviction (w/ float ms -> s at the boundary) and keep current_time for last_success_at and Failure#occurred_at. Redis unchanged, monotonic epochs don't compare across hosts.

AFAIK `CLOCK_MONOTONIC` stops during suspend, so stale events survive hibernation. If active-time behavior is acceptable I can do it as a follow-up PR with a backward-step regression spec.

bolshakov: Hey, do you think the current pr still makes sense if we gonna change wall clock time to monotonic time? It seems like the shape before the change fits more here so the SlidingWindow abstraction controls the clock used. 

Also maybe the `#sum_in_window` need to change to accept window size rather then absolute time. 

```rbs
def sum_in_window: (Integer window_size) -> Integer
```

OursCodeur: Yeah sounds right. Clock injected into SlidingWindow, duration-based sum_in_window, a backward-step regression spec, and a Timecop shim so specs and features can age the monotonic window. Or a fresh PR if you'd rather close this one, your call.

bolshakov: I think we con repurpose this pr, it's fine. Great we figured it out and found a real problem

OursCodeur: Had it stashed in advance ;) force-pushed the monotonic shape and updated the description.

bolshakov (approved):
Great job 👏

Issue #872, opened 2026-08-07
High memory usage for high throughput lights


jensljungblad-kisi:
From https://github.com/bolshakov/stoplight/discussions/871.

**Describe the bug**
When Stoplight is configured with `error_rate` and `window_size` it consumes a lot of memory regardless of window size.

**To Reproduce**
We use the following configuration
```ruby
Stoplight.configure do |config|
  config.traffic_control = :error_rate
  config.window_size = 300
  config.threshold = 0.5
  config.cool_off_time = 60
  config.recovery_threshold = 5
end

Stoplight seems to keep data about each individual call in Redis for 1 day.
```

**Expected behavior**
Data that is no longer relevant should be kept in Redis.

**Environment (please complete the following information):**
 - Ruby 4.0.3
 - Version 5.8.2

bolshakov:
Fixed in https://github.com/bolshakov/stoplight/releases/tag/v5.8.3

ildarkayumov:
Hi @bolshakov 
Thank you for your quick fix!
It seems like the issue still exists, but it is capped by a 1-hour bucket size now:

```ruby
redis = Redis.new
light = Stoplight("Example")

loop do 
  sleep 0.5

  light.run { 1 / 1 }
  key = "stoplight:v5:metrics:#{light.name}:success:#{Time.current.beginning_of_hour.to_i}"
  ttl = redis.ttl(key)
  total = redis.memory(:usage, key)
  count = redis.zrange(key, 0, -1).size
  size = total / count

  puts "#{Time.current}: ttl: #{ttl}, size: #{size}, count: #{count}, total: #{total}"
end
```

> 2026-08-27 13:41:51 UTC: ttl: 300, size: 152, count: 1, total: 152
2026-08-27 13:41:52 UTC: ttl: 300, size: 104, count: 2, total: 208
2026-08-27 13:41:52 UTC: ttl: 300, size: 85, count: 3, total: 256
2026-08-27 13:41:53 UTC: ttl: 300, size: 80, count: 4, total: 320
2026-08-27 13:41:53 UTC: ttl: 300, size: 70, count: 5, total: 352
2026-08-27 13:41:54 UTC: ttl: 300, size: 69, count: 6, total: 416
2026-08-27 13:41:54 UTC: ttl: 300, size: 68, count: 7, total: 480
2026-08-27 13:41:55 UTC: ttl: 300, size: 68, count: 8, total: 544
[...]

PR #874, opened 2026-08-08
Reduce space consumed by windowed metrics

bolshakov:
Fixes #872 

### Summary 

Fix PR reduce Redis memory consumption for error rate strategy.

### Background 

Under the hood we use 1 hour bucketed ZSETs to store metrics. For example:

- `metrics.{light_name}.success:1786176000`
- `metrics.{light_name}.success:1786179600`
- `metrics.{light_name}.success:1786183200`

Each bucket stores a full list of requests performed during this 1 hour window. This allows performing very precise calculations as cardinality of every bucket could be calculated with `O(log(N))` time complexity, yet accessing a few Redis keys per call (typically 1-2).

This comes with the memory cost - we store every request attempt as a 12-char string and memory could accumulate really fast on a busy light. 

### Solution 

Before that fix windowed metrics' memory consumption scaled lineraly with number of request regardless the used window. This fix expire stored metrics when window itself expire making space consumption proportional to window_size.

**Note**: the problem is already fixed in the develop branch, but since develop accumulated breaking changes, this PR targets main directly.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    stoplight (5.8.2)
+    stoplight (5.8.3)
       concurrent-ruby
       zeitwerk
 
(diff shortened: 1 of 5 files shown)


Lokideos (approved):
🚀

bolshakov:
I merged this manually, as develop changed drastically from main

PR #877, opened 2026-08-16
Fix state snapshot yellow color computation logic

### LINKED ISSUE #782: Two different notions of "recovery has started," one time-guarded, one not
## Background

A Stoplight circuit breaker (`Light`) is always one of three colors:

- **GREEN** -- healthy, requests run normally.
- **RED** -- tripped. Requests are blocked (or a fallback runs) without executing the
  wrapped code.
- **YELLOW** -- "recovery" / half-open. One request is allowed through as a probe; if
  it succeeds the light goes GREEN, if it fails it goes back to RED.

The color isn't stored directly. It's computed by `StateSnapshot#color`
(`lib/stoplight/domain/state_snapshot.rb:27`) from a handful of stored timestamps
(`breached_at`, `recovery_scheduled_after`, `recovery_started_at`) plus the time the
snapshot was taken (`time`). This issue is about the part of `#color` that decides
YELLOW.

There are two distinct ways a light becomes YELLOW, and `#color` has to detect both:

1. **Implicitly, by timeout.** When a light trips to RED, the store records
   `recovery_scheduled_after = now + cool_off_time` -- e.g. "this light may attempt
   recovery starting at 10:31:00." No timer or background job runs at 10:31:00; instead,
   *every* call to `Light#run` recomputes the color from scratch. Once the wall clock
   passes `recovery_scheduled_after`, `#color` starts reporting YELLOW purely by
   comparing two timestamps in the snapshot, before anything is written to the store.
   That's what routes the next call into `YellowRunStrategy`, whose `enter_recovery`
   (`lib/stoplight/domain/strategies/yellow_run_strategy.rb:123`) then does the real
   work: writes `recovery_started_at`, clears metrics, and notifies.
2. **Explicitly, once recovery has actually started.** After step 1 happens once,
   `recovery_started_at` is set (to the current time, never a future time) and
   `recovery_scheduled_after` is cleared. From then on, `#color` needs a different
   signal for YELLOW, because `recovery_scheduled_after` is gone.

So `#color`'s YELLOW branch is an `||` of two checks, one for each path. **They are not
redundant with each other** -- don't "simplify" this into a single check. Only the
second one is the bug described below.

## Location

`lib/stoplight/domain/state_snapshot.rb:32` (the `#color` method) vs.
`lib/stoplight/domain/state_snapshot.rb:46-52` (`#recovery_started?`)

```ruby
elsif (recovery_scheduled_after && recovery_scheduled_after! < time) || recovery_started_at
  Color::YELLOW
```

## The bug

The second check, `recovery_started_at`, only asks "is this field set at all?" (truthy
check). But elsewhere in the same file, `#recovery_started?` answers what sounds like
the same question more strictly:

```ruby
def recovery_started?
  if recovery_started_at.nil?
    false
  else
    recovery_started_at! <= time
  end
end
```

This is used by `YellowRunStrategy#enter_recovery` to decide whether some other thread
already kicked off recovery, so this one shouldn't do it again. It requires not just
that `recovery_started_at` is set, but that it's `<= time` (i.e., not in the future).

Today, `#color`'s bare truthy check and `#recovery_started?`'s stricter check always
agree, because both `recovery_started_at` and `time` are read from the *same* stored
snapshot, populated by the *same* clock (see `infrastructure/memory/storage/state.rb`
and `infrastructure/redis/storage/state.rb`). So this isn't a live bug -- nothing is
broken right now.

It's a latent trap, though: nothing documents that `#color` and `#recovery_started?`
are expected to answer the same question, and nothing stops a future storage backend
from deriving `time` and `recovery_started_at` from different clocks (plausible in a
distributed Redis deployment where app hosts and Redis have independent clocks). If that
ever happens, `#color` and `#recovery_started?` could disagree about whether recovery
has started, for the same snapshot.

## Suggested fix

In `#color`, replace the bare `recovery_started_at` truthy check with a call to
`#recovery_started?`, so there is exactly one place that defines "has recovery
explicitly started":

```ruby
elsif (recovery_scheduled_after && recovery_scheduled_after! < time) || recovery_started?
  Color::YELLOW
```

Leave the first check (`recovery_scheduled_after! < time`) exactly as it is -- it's the
timeout-based path described above, and it isn't part of this bug.

Make sure to reproduce this problem in a test first.

nebiyuelias1:
Fix discrepancy in logic found in the way color calculates yellow. Previously it checks whether or not  `recovery_started_at` is truthy or not. But there is another function `recovery_started?` that has additional logic of checking whether or not `recovery_started_at` is in the future or not. Therefore, the color computation also uses this function to avoid drift further down the line.

Fixes #782

### DIFF
--- a/lib/stoplight/domain/state_snapshot.rb
+++ b/lib/stoplight/domain/state_snapshot.rb
@@ -29,7 +29,7 @@ module Stoplight
           Color::GREEN
         elsif locked_state == State::LOCKED_RED
           Color::RED
-        elsif (recovery_scheduled_after && recovery_scheduled_after! < time) || recovery_started_at
+        elsif (recovery_scheduled_after && recovery_scheduled_after! < time) || recovery_started?
           Color::YELLOW
         elsif breached_at
           Color::RED
--- a/lib/stoplight/domain/state_snapshot.rb
+++ b/lib/stoplight/domain/state_snapshot.rb
@@ -44,11 +44,7 @@ module Stoplight
       # This method indicates whether the recovery has already started explicitly
       #
       def recovery_started?
-        if recovery_started_at.nil?
-          false
-        else
-          recovery_started_at! <= time
-        end
+        !!recovery_started_at
       end
 
       def recovery_started_at!
--- a/spec/unit/stoplight/domain/state_snapshot_spec.rb
+++ b/spec/unit/stoplight/domain/state_snapshot_spec.rb
@@ -46,17 +46,17 @@ RSpec.describe Stoplight::Domain::StateSnapshot do
       it { is_expected.to be(Stoplight::Color::YELLOW) }
     end
 
-    context "when threshold is breached and recovery is in the future" do
-      let(:breached_at) { time - 3 }
-      let(:recovery_started_at) { time + 3 }
-
-      it { is_expected.to be(Stoplight::Color::RED) }
-    end
 
     context "when threshold breached" do
       let(:breached_at) { time - 3 }
 
       it { is_expected.to be(Stoplight::Color::RED) }
+
+      context "when recovery is in the future" do
+        let(:recovery_started_at) { time + 3 }
+
+        it { is_expected.to be(Stoplight::Color::RED) }
+      end
     end
   end
 end
--- a/spec/unit/stoplight/domain/state_snapshot_spec.rb
+++ b/spec/unit/stoplight/domain/state_snapshot_spec.rb
@@ -46,6 +46,12 @@ RSpec.describe Stoplight::Domain::StateSnapshot do
       it { is_expected.to be(Stoplight::Color::YELLOW) }
     end
 
+    context "when recovery is in the future due to clock skew" do
+      let(:recovery_started_at) { time + 3 }
+
+      it { is_expected.to be(Stoplight::Color::YELLOW) }
+    end
+
     context "when threshold breached" do
       let(:breached_at) { time - 3 }
 
--- a/spec/unit/stoplight/domain/state_snapshot_spec.rb
+++ b/spec/unit/stoplight/domain/state_snapshot_spec.rb
@@ -46,6 +46,13 @@ RSpec.describe Stoplight::Domain::StateSnapshot do
       it { is_expected.to be(Stoplight::Color::YELLOW) }
     end
 
+    context "when threshold is breached and recovery is in the future" do
+      let(:breached_at) { time - 3 }
+      let(:recovery_started_at) { time + 3 }
+
+      it { is_expected.to be(Stoplight::Color::RED) }
+    end
+
     context "when threshold breached" do
       let(:breached_at) { time - 3 }
 
--- a/spec/unit/stoplight/domain/state_snapshot_spec.rb
+++ b/spec/unit/stoplight/domain/state_snapshot_spec.rb
@@ -46,17 +46,10 @@ RSpec.describe Stoplight::Domain::StateSnapshot do
       it { is_expected.to be(Stoplight::Color::YELLOW) }
     end
 
-
     context "when threshold breached" do
       let(:breached_at) { time - 3 }
 
       it { is_expected.to be(Stoplight::Color::RED) }
-
-      context "when recovery is in the future" do
-        let(:recovery_started_at) { time + 3 }
-
-        it { is_expected.to be(Stoplight::Color::RED) }
-      end
     end
   end
 end


bolshakov:
Please take a look at the tests - something is broken and we need to figure out if this behavior change is intentional or not

spec/unit/stoplight/domain/state_snapshot_spec.rb
```diff
@@ -46,6 +46,13 @@
       it { is_expected.to be(Stoplight::Color::YELLOW) }
     end
 
+    context "when threshold is breached and recovery is in the future" do
```
Lokideos: I think we should nest this spec context inside the `when threshold breached` context below like this:

```ruby
#...
context "when threshold breached" do
  let(:breached_at) { time - 3 }

  it { is_expected.to be(Stoplight::Color::RED) }

  context "when recovery is in the future" do
    let(:recovery_started_at) { time + 3 }

    it { is_expected.to be(Stoplight::Color::RED) }
  end
end
```

WDYT?

lib/stoplight/domain/state_snapshot.rb
```diff
@@ -29,7 +29,7 @@ def color
           Color::GREEN
         elsif locked_state == State::LOCKED_RED
           Color::RED
-        elsif (recovery_scheduled_after && recovery_scheduled_after! < time) || recovery_started_at
+        elsif (recovery_scheduled_after && recovery_scheduled_after! < time) || recovery_started?
```
Lokideos, line 32: I think we cannot realistically reach state in which we have set `breached_at` and `recovery_started_at` since we clear metrics across data stores at the moment when we transition to yellow (when we set `recovery_started_at`):
- [Memory Data Store](https://github.com/nebiyuelias1/stoplight/blob/ae7ab15efbe56a37e59447f8a4a48e7dfa7eb362/lib/stoplight/infrastructure/memory/storage/state.rb#L115-L130)
- [Redis Data Store](https://github.com/nebiyuelias1/stoplight/blob/6c889fdca69e7418079e0588d2a2a8a774c1d3db/lib/stoplight/infrastructure/redis/storage/state/transition_to_yellow.lua)

Which leads to an interesting case in which in the event of clocks skew we will return `GREEN` color since `brached_at` is `nil`, but we should return `RED`. 🤔 

@bolshakov, I think that from the CB logic percpective it means that if we either have `breached_at` set or `recovery_started_at` set in the future it means that the Color is `RED` (meaning we should slightly adjust logic for Red color below).
WDYT?

Lokideos (commented):
Great first issue! 🚀  Keep on keeping on ^_^
Before merging:
1. We should investigate why specs are failing here and make sure that they are green.
2. Address incorrect Light state in the event of clock skew (we should take into account existance of both `breached_at` and `recovery_started_at` fields).

WDYT? 🙂

spec/unit/stoplight/domain/state_snapshot_spec.rb
nebiyuelias1: Yeah much better this way thanks!

lib/stoplight/domain/state_snapshot.rb
bolshakov: > it means that if we either have breached_at set or recovery_started_at set in the future it means that the Color is RED (meaning we should slightly adjust logic for Red color below).

if `breached_at` is in the future , yes this indicates started recovery with clock skew probably.  But if `recovery_started_at` is in the future, it means that:
- another process started recovery
- two process has clock skew related to each other 

I don't think we should treat such case as red. This would lead to two process reaching different decisions - one already declared recovery and another one thinks it's red.

Lokideos: But what if `breached_at` is `nil`, since we clear the metrics during recovery attempt and `recovery_started_at` is in the future? 🤔

bolshakov: Also indicates clock skew. We have two separate fields - "recovery scheduled after" which is set in the future when we enter red state and "recovery started at" which us set when we enter yellow state.

So the first field allows all instances align when recovery should start and the second one helps them get o learn that recovery has already started (most likely the first prove is about to fire or already fired)

spec/unit/stoplight/domain/state_snapshot_spec.rb
nebiyuelias1: @Lokideos I just realized technically breached_at and recovery_started_at cannot have not nil values at the same time (according to [this](https://github.com/bolshakov/stoplight/pull/877#discussion_r3795329261)), so is nesting a good idea in this case?

lib/stoplight/domain/state_snapshot.rb
nebiyuelias1: So @bolshakov are you suggesting it's okay if the color is Green in this case?

spec/unit/stoplight/domain/state_snapshot_spec.rb
Lokideos: Totally, let's move it outside of the context w/o any `breached_at` setup (it should be `nil`).

Lokideos (commented):
@nebiyuelias1, after reviewing the issue, I now know how to circumvent the potential clock skew. We still should rely on `#recovery_started?` method rather than `recovery_started_at` field, but the method itself should be changed.
In `#recovery_started?` we check if the `recovery_started_at` field exists and if it's in the past.
But for making a decision if the Stoplight is in recovery, we do not need to know if `recovery_started_at` is in the past or in the future. It should just be set. Meaning that `#recovery_started?` method should just check the existence of the field and be changed to:

```ruby
def recovery_started?
  !!recovery_started_at
end
```

This will allow us to prevent any potential time skew issues (since we do not compare time at all here) and still have a timestmap for recovery-related metrics.
WDYT? 🙂

Lokideos (approved):
Looks great 🚀

PR #879, opened 2026-08-18
Wire RecoveryProbeCompleted event (#746)

### LINKED ISSUE #746: Wire `RecoveryProbeCompleted` event

The original design is PR #721,  -- read its
`RecoveryProbeCompleted`.

## What you're wiring up, and why it isn't `RunCompleted`

A light's `#run` method behaves differently depending on its color. When it's **yellow**
("half-open" -- the circuit is cautiously letting one probe request through to test whether the
dependency has recovered), each attempt is a *recovery probe*, not an ordinary run.
`RecoveryProbeCompleted` (`lib/stoplight/domain/telemetry/recovery_probe_completed.rb`,
`sig/stoplight/telemetry/recovery_probe_completed.rbs`) is the yellow-specific sibling of
`RunCompleted` -- it exists precisely so a probe's outcome doesn't get mixed into general
error-rate metrics alongside real traffic. This split was already decided when `RunCompleted`
was wired up: `YellowRunStrategy` (`lib/stoplight/domain/strategies/yellow_run_strategy.rb`) was
deliberately left untouched by that work, and should stay untouched by `RunCompleted` here too --
this ticket gives it its own event instead.

## Where to make the change

`Domain::Tracker::RecoveryProbe#record_failure` / `#record_success`

```ruby
def record_failure(exception)
  metrics_store.record_failure(exception)
  recover
  # emit RecoveryProbeCompleted(outcome: :failure, ...)
end

def record_success
  metrics_store.record_success
  recover
  # emit RecoveryProbeCompleted(outcome: :success, ...)
end
```

Unlike the recovery-transition events in the other tickets in this set, this one fires **on
every single probe**, unconditionally -- there's no dedup guard to worry about, because it's not
reporting a state transition, just "a probe happened."

## Payload fields

- `outcome:` -- `:success` or `:failure` (check the exact allowed values against the
  `probe_outcome` type in `sig/stoplight/telemetry/recovery_probe_completed.rbs`).
- `duration_ms:` -- the probe's latency. The catch: `Tracker::RecoveryProbe` doesn't know timing
  at all -- the actual call boundary is `YellowRunStrategy#execute`, which already has clock
  access from the `RunCompleted` work. You'll need to measure duration there and pass it down
  into `record_recovery_probe_success` / `record_recovery_probe_failure`, which currently take
  no duration argument.
- `failure:` -- `Domain::Telemetry::Failure.new(exception:, tracked: true)` on the failure path,
  `nil` on success.
- `progress:` -- recovery metrics *after* this probe. `#recover` already computes
  `recovery_metrics = metrics_store.metrics_snapshot`, and since `record_failure`/`record_success`
  update `metrics_store` before calling `#recover`, that snapshot should already reflect this
  probe -- worth double-checking with a test rather than assuming.

## Suggested approach

1. Since timing has to happen in `YellowRunStrategy`, this may want its own small
   duration-measuring collaborator, similar in spirit to `run_recorder.rb` (built for
   `RunCompleted`) but not necessarily the same shape -- decide what fits once you see the call
   site.
2. Add spec coverage both in
   `spec/unit/stoplight/domain/run_strategies/yellow_run_strategy_spec.rb` (duration is measured)
   and `spec/unit/stoplight/domain/tracker/recovery_probe_spec.rb` (event fires with the right
   outcome/failure/progress).

## References

- `sig/stoplight/telemetry/recovery_probe_completed.rbs`

ChrisJr404:
Wires up `RecoveryProbeCompleted`, which was defined back in the telemetry work but never actually emitted, so subscribers can see every yellow probe's outcome without it getting mixed into the general `RunCompleted` error-rate stream.

It fires from `RecoveryProbe#recover`, right after the post-probe metrics snapshot and before the still-yellow early return, so it happens on every probe regardless of which color the probe resolves to. There's no dedup guard because it's a per-probe measurement, not a state transition. `record_success`/`record_failure` pass the outcome (`:success`/`:failure`), the tracked failure (nil on success), and the same snapshot used for the recovery decision as `progress`.

Duration is the one thing `RecoveryProbe` can't see on its own, so `YellowRunStrategy` measures it (it already has clock access from the `RunCompleted` work) and threads it down. I typed `duration_ms` as `Float?` to match `RunCompleted#duration_ms` — it's nil only when nobody was subscribed at probe start, same race. I also extended the clock-capture guard so it measures when the probe event is subscribed, not just `RunCompleted`; otherwise subscribing to only `RecoveryProbeCompleted` would always give you a nil duration.

Tests: added a `RecoveryProbeCompleted telemetry` group in `recovery_probe_spec.rb` covering success, tracked failure, that it fires even when the light stays yellow, and that a nil duration passes through. In `yellow_run_strategy_spec.rb` I added cases for the measured duration reaching the probe and for the probe-only-subscribed path, and updated the existing expectations for the new `duration_ms:` argument. RBS updated in lockstep.

`bundle exec rspec` is green (791 examples, 0 failures, with Redis up), `steep check` clean, `rbs -I sig validate` clean, `standardrb` clean.

Closes #746.

### DIFF
--- a/lib/stoplight/domain/strategies/yellow_run_strategy.rb
+++ b/lib/stoplight/domain/strategies/yellow_run_strategy.rb
@@ -104,7 +104,7 @@ module Stoplight
         end
 
         def capture_started_at
-          @clock.monotonic_time if @run_recorder.subscribed?
+          @clock.monotonic_time if @run_recorder.subscribed? || request_tracker.subscribed?
         end
 
         def duration_since(started_at)
@@ -113,12 +113,12 @@ module Stoplight
 
         def record_recovery_probe_success(duration_ms:, error: nil)
           @run_recorder.record_success(duration_ms:, error:)
-          request_tracker.record_success
+          request_tracker.record_success(duration_ms:)
         end
 
         def record_recovery_probe_failure(error, duration_ms:, fallback_used:)
           @run_recorder.record_failure(error, duration_ms:, fallback_used:)
-          request_tracker.record_failure(error)
+          request_tracker.record_failure(error, duration_ms:)
         end
 
         def enter_recovery(state_snapshot)
--- a/lib/stoplight/domain/tracker/recovery_probe.rb
+++ b/lib/stoplight/domain/tracker/recovery_probe.rb
@@ -13,17 +13,20 @@ module Stoplight
           @emitter = emitter
         end
 
+        # True when any subscriber would receive a RecoveryProbeCompleted event.
+        def subscribed? = emitter.subscribed?(Telemetry::RecoveryProbeCompleted)
+
         # @param exception [Exception]
-        def record_failure(exception)
+        def record_failure(exception, duration_ms:)
           metrics_store.record_failure(exception)
 
-          recover
+          recover(outcome: :failure, duration_ms:, failure: Telemetry::Failure.new(exception:, tracked: true))
         end
 
-        def record_success
+        def record_success(duration_ms:)
           metrics_store.record_success
 
-          recover
+          recover(outcome: :success, duration_ms:, failure: nil)
         end
 
         private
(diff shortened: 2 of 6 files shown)


ChrisJr404:
Going to withdraw this one. These `good first issue` tickets should stay open for folks getting their first contributions in — I didn't mean to jump the queue. Closing so it's free for someone who needs the on-ramp. Sorry for the noise.

PR #887, opened 2026-08-22
Prune expired Memory metric buckets on writes

### LINKED ISSUE #774: `WindowMetrics`'s success bucket in `Memory` Data Store never prunes on the healthy path, growing unbounded for the life of the process
## Summary

`SlidingWindow#increment` only adds a bucket and never evicts, and pruning only happens inside `#sum_in_window`.
`WindowMetrics#record_success` never triggers a snapshot/read, so for an error-rate light that mostly succeeds --
the normal, desired state -- the success bucket hash accumulates one entry per unique second forever, with
nothing to cap it.

## Location

`lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb:28-31` (`#increment`),
`lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb:33-36` (`#sum_in_window`, the only
place that prunes via `slide_window!`),
`lib/stoplight/infrastructure/memory/storage/window_metrics.rb:46-59` (`record_success`)

## Details

`window_metrics.rb`'s `record_success` never calls `metrics_snapshot`/`sum_in_window`, and upstream,
`domain/tracker/request.rb:28`'s `Request#record_success` never asks for a snapshot either. For a light using the
error-rate (`WindowMetrics`) strategy that mostly succeeds, `@successes.buckets` accumulates one hash entry per
unique second, forever, for the lifetime of the process, with nothing to cap it.

Contrast with the Redis backend for the same store (`window_metrics/record_success.lua`), which sets `EXPIRE` on
every write independent of reads, bounding memory by construction. The Memory backend has no equivalent
write-time bound.

This is unobservable today: nothing logs or measures bucket count, so it can't be detected before it manifests as
OOM/GC pressure on a long-lived process. Severity is High for any long-lived process using the error-rate
strategy -- unbounded resource growth with no backpressure and no observability into the growth.

## Suggested fix

Prune on write too (slide the window inside `increment`, not only inside `sum_in_window`), or cap bucket
count/age unconditionally on every mutation, mirroring the Redis backend's per-write TTL approach.

Make sure amortized O(1) property is respected.

henriquejsza:
## Summary

- prune expired in-memory metric buckets during each write
- keep the configured window inside `SlidingWindow` while preserving one monotonic-clock read per increment
- cover sustained successful traffic that never requests a metrics snapshot

Each bucket is still removed from the FIFO hash at most once, so writes remain amortized O(1) while memory stays bounded to the active window.

## Validation

- `bundle exec rspec` — 796 examples, 0 failures
- `bundle exec standardrb`
- `bundle exec steep check`
- `bundle exec rbs -I sig validate`
- Cucumber with Memory and Redis using both `Stoplight()` and `System#register`
- 1,000,001-write stress smoke — 11 buckets retained for a 10-second window

Closes #774

### DIFF
--- a/lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb
+++ b/lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb
@@ -35,8 +35,8 @@ module Stoplight
               @running_sum += 1
             end
 
-            def sum_in_window(window_size)
-              slide_window!(monotonic_seconds - window_size)
+            def sum_in_window
+              slide_window!(monotonic_seconds - @window_size)
               @running_sum
             end
 
--- a/lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb
+++ b/lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb
@@ -18,17 +18,20 @@ module Stoplight
           # @note Not thread-safe; synchronization must be handled externally
           # @api private
           class SlidingWindow
-            def initialize(clock:)
+            def initialize(clock:, window_size:)
               # A hash mapping time buckets to their counts
               @buckets = Hash.new { |buckets, bucket| buckets[bucket] = 0 }
               # The running sum of all increments in the current window
               @running_sum = 0
               @clock = clock
+              @window_size = window_size
             end
 
             # Increment the count at the current monotonic second
             def increment
-              @buckets[current_bucket] += 1
+              timestamp = monotonic_seconds
+              slide_window!(timestamp - @window_size)
+              @buckets[timestamp.to_i] += 1
               @running_sum += 1
             end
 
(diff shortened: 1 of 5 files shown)


lib/stoplight/infrastructure/memory/storage/window_metrics/sliding_window.rb
bolshakov: Could you update the method signature to remove this parameter? Now we already pass window size to the constructor and method parameter seems redundant

henriquejsza: Updated — `sum_in_window` now uses the `window_size` configured on construction. I also updated both callers, the RBS signature, and the direct specs.

bolshakov (approved):

bolshakov:
Great job!

PR #893, opened 2026-08-24
Fix/adjust monotonic time functions

### LINKED ISSUE #880: Add `monotonic_seconds` to the clock interface and rename `monotonic_time`
## Problem

`SlidingWindow` exposes a private `monotonic_seconds` helper that converts `@clock.monotonic_time` (ms) to seconds:

```ruby
def monotonic_seconds
  @clock.monotonic_time / 1000.0
end
```

The ms→s division is a unit-conversion concern that belongs on the clock, not on every caller. `monotonic_time` also has an ambiguous name — the unit is not apparent from the method name.

## Proposed change

- Add `monotonic_seconds` to the clock interface (`SystemClock`, `NullClock`) returning `Float` seconds via `Process.clock_gettime(Process::CLOCK_MONOTONIC, :float_second)`
- Rename `monotonic_time` → `monotonic_millis` to make the unit explicit
- Update `SlidingWindow` to call `@clock.monotonic_seconds` directly, removing its own conversion helper
- Update RBS signatures for both clock types and `SlidingWindow`

## Acceptance criteria

- `SystemClock#monotonic_seconds` returns float seconds
- `SystemClock#monotonic_millis` replaces `monotonic_time`
- `NullClock` exposes both methods
- `SlidingWindow` removes its private conversion helper and calls `@clock.monotonic_seconds`
- RBS updated throughout
- `system_clock_spec` covers both new methods

nebiyuelias1:
Closes #880

### DIFF
--- a/lib/stoplight/domain/strategies/green_run_strategy.rb
+++ b/lib/stoplight/domain/strategies/green_run_strategy.rb
@@ -53,11 +53,11 @@ module Stoplight
         attr_reader :request_tracker
 
         def capture_started_at
-          @clock.monotonic_time if @run_recorder.subscribed?
+          @clock.monotonic_millis if @run_recorder.subscribed?
         end
 
         def duration_since(started_at)
-          @clock.monotonic_time - started_at if started_at
+          @clock.monotonic_millis - started_at if started_at
         end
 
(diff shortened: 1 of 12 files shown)


bolshakov:
Great work, thank you!  🚀

PR #898, opened 2026-08-27
Redis Window Metrics: redesign

bolshakov:
Addresses #872 

## How we got here

**2022 -- original design.** Redis stored failures in a single ZSET per light. `record_failure` called `ZADD` + `ZREMRANGEBYSCORE` (trim by time) + `ZREMRANGEBYRANK` (cap at threshold). `record_success` wiped the entire ZSET. Memory was bounded to `threshold` failures at most.

This design only tracked failures, and it was correct for the one strategy that existed: **consecutive failures**. Clearing on success is the right semantics when you only care about the current streak.

**May 2025 -- error-rate strategy (see #313).** The new `error_rate` traffic control requires both successes and failures to be tracked persistently -- clearing either set on the other event would corrupt the ratio. The old single-ZSET approach was
broken for this use case, storage was redesigned from scratch.

The new design introduced **time-bucketed ZSETs**: fixed 1-hour buckets keyed by `{metric}:{bucket_timestamp}`. Metadata (consecutive counters, last error) moved to a separate hash. 

Benchmarked bucket sizes and aggregate hash-counter approaches. 3600s buckets won on throughput, we chose them. **No benchmark compared buckets against two independent ZSETs with `ZREMRANGEBYSCORE` on write** -- that option was never evaluated because it had been discarded in its old (broken-for-error-rate) form without being reconsidered in its corrected form.

The hash-conture approach was ruled out in favor of current ZSET approach as hash-counters degrade with window size.

## The problem

With 1-hour buckets and `window_size = 300s`, the bucket TTL is `300 + 3600 = 3900s`. Every event written in the current bucket stays in Redis for up to one hour regardless of the window size. At 2 req/s with a 5-minute window:

- **Events a user cares about**: 300s x 2 = 600
- **Events actually stored**: up to 3600s x 2 = 7_200 (12x)

At 50 req/s with 100 lights, that is ~30 MB per light pair (success + failure) -- or ~3 GB for the fleet -- for data that is 95% outside the query window.

## Solution 

I built an  implementation with two ZSETs and `ZREMRANGEBYSCORE` on every write. It passes all shared unit test examples and all end-to-end Cucumber scenarios. 

### Memory

| window_size | rate (r/s) | bucketed | two-ZSET | ratio    |
|-------------|------------|----------|----------|----------|
| 60 s        | 2          | 1.34 MB  | 6.5 KB   | **211x** |
| 60 s        | 50         | 30.7 MB  | 353 KB   | **89x**  |
| 300 s       | 2          | 1.33 MB  | 71 KB    | **19x**  |
| 300 s       | 10         | 6.54 MB  | 352 KB   | **19x**  |
| 300 s       | 50         | 29.2 MB  | 1.68 MB  | **17x**  |
| 3600 s      | 2          | 1.32 MB  | 837 KB   | 1.6x     |
| 3600 s      | 50         | 32.4 MB  | 20.7 MB  | 1.6x     |

The bucketed approach retains `bucket_size x rate` events per bucket (up to two alive at once). The two-ZSET approach retains exactly `window_size x rate` events -- the theoretical minimum for per-event storage with exact sliding-window semantics.

For `window_size = 3600 s` (equal to `bucket_size`) the approaches converge at ~1.6x (two full buckets vs one full ZSET).

### Throughput

| window_size | bucketed        | two-ZSET        | delta        |
|-------------|-----------------|-----------------|--------------|
| 60 s        | 5 700–6 300 i/s | 5 900–6 600 i/s | ≤ 5 % faster |
| 300 s       | 5 900–6 300 i/s | 6 400–6 500 i/s | ~7 % faster  |
| 3600 s      | 5 900–6 300 i/s | 6 200–6 750 i/s | ~7 % faster  |

Two-ZSET is consistently faster or equal across all scenarios. The reason: at steady state the ZSET contains at most `window_size x rate` entries, while the bucket contains up to `bucket_size x rate` entries. Since `bucket_size ≥ window_size` for all small windows, the bucket is always at least as large -- and O(log N) ZSET amortized operations (`ZADD`, `ZCOUNT`, `ZREMRANGEBYSCORE`) are faster on the smaller set.

## Proposed design for v6.0

Replace `WindowMetrics` with two ZSETs and `ZREMRANGEBYSCORE` on write.

### Storage structure

```
{key_space}:window_metrics            -- hash, metadata (consecutive counters, last error, seq)
{key_space}:window_metrics:success    -- ZSET, one member per success, score = timestamp
{key_space}:window_metrics:failure    -- ZSET, one member per failure, score = timestamp
```

Three keys per light (down from five with two-bucket steady state).

### Write path (Lua, atomic)

`record_success`:

1. `HINCRBY metadata_key seq 1` -- get next sequence number
2. `ZADD success_key timestamp seq`
3. `ZREMRANGEBYSCORE success_key -inf (window_start_ts` -- prune events outside window
4. `EXPIRE success_key window_size * 2`
5. Update metadata hash (consecutive counters, `last_success_at`)

`record_failure` (also returns snapshot in one round trip):

1. `HINCRBY metadata_key seq 1` -- get next sequence number
2. `ZADD failure_key timestamp seq`
3. `ZREMRANGEBYSCORE failure_key -inf (window_start_ts`
4. `EXPIRE failure_key window_size * 2`
5. Update metadata hash
6. `ZCOUNT success_key window_start_ts timestamp` -- read for snapshot
7. `ZCOUNT failure_key window_start_ts timestamp` -- read for snapshot

### Read path

`metrics_snapshot`:

1. `ZCOUNT success_key window_start_ts window_end_ts`
2. `ZCOUNT failure_key window_start_ts window_end_ts`
3. `HMGET metadata_key ...fields`

All three calls execute in one Lua script (one round trip).

### Breaking change

Key layout changes: the bucket-timestamp suffix is dropped. Existing bucket keys expire naturally on their current TTL. Fine under a major-version release.

### Next steps

- further reduce storage requirements by using monotonically raising request ids 
- implement sampling rate


bolshakov:
Moving this to draft as I might have found a more memory-efficient approach

bolshakov:
Replaced by https://github.com/bolshakov/stoplight/pull/904

PR #900, opened 2026-08-27
Redis Window Metrics: smaller event keys

bolshakov:
Further reduces amount of data stored as ZSET members for window metrics. Benchmark results in. At 15,000 members (50 req/s, 300s window):

| approach | KB | bytes/member |
|--------|--------|--------|
| hex-24 (SecureRandom) | 1,730 KB | 118 |
| int-seq (HINCRBY) | 1,494 KB | 102 |
| savings | ~236 KB/ZSET  | ~16 | 

Performance benchmark results: difference falls withing noise


bolshakov:
Replaced by https://github.com/bolshakov/stoplight/pull/904

PR #902, opened 2026-08-28
Remove min_requests from ErrorRate public API

bolshakov:
Error Rate strategy no longer accepts a min_requests: keyword argument. The minimum sample size is now fixed at 100 requests (MIN_REQUESTS = 100).

## why

The error rate is only useful as a circuit-breaker signal when it is based on enough requests.

With a very small number of requests, a single failure can change the error rate dramatically. For example, with 10 requests, one failure changes the measured error rate by 10 percentage points. This can cause the circuit breaker to react to normal random variation rather than a genuine problem.

Requiring at least 100 requests makes the measurement more stable: each individual request changes the measured error rate by only 1 percentage point.

100 is therefore used as a fixed minimum rather than allowing callers to configure arbitrarily small values.

As a rule of thumb, statistical analysis of proportions also shows that around 100 observations gives a worst-case 95% margin of error of roughly ±10 percentage points. This is not a guarantee of accuracy, but it illustrates why very small samples provide much less reliable measurements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,18 +353,9 @@ light = Stoplight(
 
 Monitors error rate over a 5-minute sliding window. The stoplight turns red when error rate exceeds 50%.
 
-```ruby
-light = Stoplight(
-  "Payment API", 
-  traffic_control: {
-    error_rate: { min_requests: 20 },
-  }, 
-  window_size: 300, 
-  threshold: 0.5,
-)
-```
-
-Only evaluates error rate after at least 20 requests within the window. Default `min_requests` is 10.
+Error rate evaluation starts only after 100 requests within the window — enough samples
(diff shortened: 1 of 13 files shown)


Lokideos (approved):
Great work! 🚀 
There are a couple of leftovers in docs:
1. [settings.rbs](https://github.com/bolshakov/stoplight/blob/ac324d606cef2d0fab1e538dc5af4f9225e396d9/sig/stoplight/domain/telemetry/settings.rbs#L14)
2. [types.rbs](https://github.com/bolshakov/stoplight/blob/320b93c6dac82bcd3aa5212bf1b252d1789fd800/sig/stoplight/wiring/types.rbs#L3)

PR #903, opened 2026-08-29
docs: Replace state diagram with new one

Lokideos:
This PR aims to replace current state diagram in README with an SVG-based diagram.
Current Diagram is a little bit confusing, so I've decided to replace it with SVG diagram with clear transition paths between states.
Now in README it looks like this - https://github.com/bolshakov/stoplight/tree/docs-update-readme-state-diagram#core-concepts


bolshakov (approved):
Nice! Love it 😍

Issue #283, opened 2025-04-07
Implement Bucket-Based Error Rate Calculation


bolshakov:
Currently, Stoplight uses a simple threshold-based approach to circuit breaking, where failures are stored in a Redis sorted set with timestamps as scores. This design has several limitations:

1. The circuit can only react to the absolute number of errors IN A ROW, not error rates
2. For high-traffic services, the current approach is memory-intensive as it stores individual failures
3. There's no built-in way to differentiate between 5 failures in 10 requests (50% error rate) versus 5 failures in 1000 requests (0.5% error rate)

We need more adaptive circuit breaking strategies that respond to error rates rather than absolute error counts.

### Proposed Solution

Implement a bucket-based time series aggregation mechanism for tracking both successes and failures. This will:

- Allow triggering circuit breaking based on error rates (percentage) rather than absolute counts
- Provide better memory efficiency for high-traffic services
- Enable more sophisticated adaptive strategies based on time-bucketed metrics

### Data Structure

Add a new Redis data structure for time-bucketed metrics:

```
stoplight:v5:stats:{light_name}:{bucket_timestamp} → Hash
  fields:
    - success: <count>
    - failure: <count>
```

Each bucket represents a discrete time interval (e.g., 1 second) and tracks the count of successes and failures within that interval.

### Configuration API

Add configuration options

```ruby
# Consecutive strategy: Stop immediately if the service is completely down (3 consecutive 
# calls led to an error, current approach)
Stoplight('payment-critical', strategy: :consecutive, threshold: 3)
Stoplight('payment-critical', threshold: 3) # the same semantics, `consecutive` by default 

# Error rate strategy: Stop if service quality degrades significantly  (10% of requests 
# within last 60 seconds led to an error)
Stoplight('payment-quality', strategy: :error_rate, threshold: 0.1, window_size: 60)

# Absolute strategy: Limit total error volume regardless of success rate (50 errors 
# winning last 5 minutes)
Stoplight('payment-volume', strategy: :absolute, threshold: 50, window_size: 300)
```

#### Implementation Approach

1. Extend the `DataStore::Redis` class to track successes alongside failures
2. Add a new Lua script for efficiently incrementing bucket counters
3. Add error rate calculation methods that aggregate across buckets
4. Add TTL-based cleanup mechanism for automatic expiry of old buckets

### Tradeoffs and Considerations

- **Memory vs. Precision**: Bucket-based approach uses less memory but loses individual event timing
[...]

PR #904, opened 2026-08-30
More space efficient sliding window implementation for Redis store

### LINKED ISSUE #872: High memory usage for high throughput lights

bolshakov:
Fixes #872 

## How we got here

**2022 -- original design.** Redis stored failures in a single ZSET per light. `record_failure` called `ZADD` + `ZREMRANGEBYSCORE` (trim by time) + `ZREMRANGEBYRANK` (cap at threshold). `record_success` wiped the entire ZSET. Memory was bounded to `threshold` failures at most.

This design only tracked failures, and it was correct for the one strategy that existed: **consecutive failures**. Clearing on success is the right semantics when you only care about the current streak.

**May 2025 -- error-rate strategy (see #313).** The new `error_rate` traffic control requires both successes and failures to be tracked persistently -- clearing either set on the other event would corrupt the ratio. The old single-ZSET approach was broken for this use case, storage was redesigned from scratch.

The new design introduced **time-bucketed ZSETs**: fixed 1-hour buckets keyed by `{metric}:{bucket_timestamp}`. Metadata (consecutive counters, last error) moved to a separate hash. 

@sirwolfgang proposed an alternative idea: to store time-buckets in Redis hash (`{bucket}:failure => count` and `{bucket}:success => count`) and an additional Sorted Set index, so we could know which buckets to sum (see https://github.com/bolshakov/stoplight/issues/283#issuecomment-2849373776)

We benchmarked bucket sizes and aggregate hash-counter approaches. 3600s buckets won on throughput, we chose them. The hash-conture approach was ruled out in favor of current ZSET approach as hash-counters degrade with window size (more and more keys to summarize)

## The problem

With 1-hour buckets and `window_size = 300s`, the bucket TTL is `300 + 3600 = 3900s`. Every event written in the current bucket stays in Redis for up to one hour regardless of the window size. At 2 req/s with a 5-minute window:

- **Events a user cares about**: 300s x 2 = 600
- **Events actually stored**: up to 3600s x 2 = 7_200 (12x)

At 50 req/s with 100 lights, that is ~30 MB per light pair (success + failure) -- or ~3 GB for the fleet -- for data that is 95% outside the query window. The memory consumption is proportional to request rate.

## Solution 

I expanded @sirwolfgang's idea with a few improvements: 
- instead of summarizing buckets in a window on every request, we maintain a running sum counter for failures and successes. 
- we use sorted set index to find buckets eligible for eviction 

This algorithm allows O(1) amortized operations complexity and O(K) memory usage (where K is window size). 

### Bencharks 

First, memory. 

1. Currently, Redis consumers space proportional to request rate, 120 B per each request and retains 2 hours or more such records.
2. This branch consumes space proportionally to a number of seconds in a window, ~300 B per *active second*, flat in request rate. Identical at 10/s and 100/s.

Absolute numbers for comparison (per light):

| window | rate/s | events in window | develop    | hash-buckets |
|--------|--------|------------------|------------|--------------|
| 1m     | 10     | 600              | 74.9 KB    | **11.6 KB**  |
| 1m     | 100    | 6,000            | 725.3 KB   | **11.7 KB**  |
| 3m     | 10     | 1,800            | 204.3 KB   | **53.3 KB**  |
| 3m     | 100    | 18,000           | 2,092.6 KB | **51.7 KB**  |
| 5m     | 10     | 3,000            | 371.3 KB   | **88.0 KB**  |
| 5m     | 100    | 30,000           | 3,356.0 KB | **88.0 KB**  |
| 10m    | 10     | 6,000            | 753.5 KB   | **172.6 KB** |
| 10m    | 100    | 60,000           | 6,672.1 KB | **174.6 KB** |

Performance-wise:
- Develop - Log(N) + M (N - number of requests, M - window size) 
- Hash-buckets - O(1) amortized (see above)

In practice difference is within measurement error. 

### Breaking change

Key layout changes: the bucket-timestamp suffix is dropped. Existing bucket keys expire naturally on their current TTL. Fine under a major-version release.


lib/stoplight/infrastructure/redis/storage/window_metrics/_evict.lua
```diff
@@ -0,0 +1,34 @@
...
+    return
+  end
+
+  local fields = {}
+  for _, bucket in ipairs(buckets) do
+    fields[#fields + 1] = bucket .. ':f'
+    fields[#fields + 1] = bucket .. ':s'
+  end
+
+  local values = redis.call('HMGET', key, unpack(fields))
```
bolshakov: unpack explodes on 8000 elements on all Redis/Valkey versions, making 4000 seconds a maximum possible window. 

This must be fixed before merging. But I decided to keep it as is at this stage to make review easier.

Lokideos (approved):
Phenomenal work! 🚀 
I have a concern that since we now evict buckets on each call, we are now uniquly vulnarable to clock skew. Since time is passed to the Redis from the Client, client with significant clock skew could always evict buckets, thus cutting the window size (or in a case of severe clock skew we will evict full buckets on each call).
What about using time on Redis server for buckets eviction at least? I think we can set it with something like:
```lua
local server_ts = tonumber(redis.call('TIME')[1])
```

bolshakov:
You're right. Previously we needed a to use ruby clock because we used them to generate Redis key names that we have to pass externally to a lua script. Now this is not an issue and we e can use redis clock. 

The problem is that testing Redis time is much harder then ruby time and I don't want to sacrifice coverage.

A few ideas we can explore:
- passing time from ruby in teats and relying on lua time in prod
- injecting time function that could be stubbed in tests

I suggest not to bring this bigger change to this PR and handle it separately as a follow up

Lokideos:
> You're right. Previously we needed a to use ruby clock because we used them to generate Redis key names that we have to pass externally to a lua script. Now this is not an issue and we e can use redis clock.
> 
> The problem is that testing Redis time is much harder then ruby time and I don't want to sacrifice coverage.
> 
> A few ideas we can explore:
> 
> * passing time from ruby in teats and relying on lua time in prod
> * injecting time function that could be stubbed in tests
> 
> I suggest not to bring this bigger change to this PR and handle it separately as a follow up

I agree, let's handle this issue in a separate PR. I think going with time function injections looks like a correct choice here 🤔

PR #910, opened 2026-08-31
Route legacy notifiers through the telemetry bus instead of domain code

bolshakov:
## Summary
- `Domain::Light`'s collaborators (`Tracker::Request`, `Tracker::RecoveryProbe`, `YellowRunStrategy`) used to call the legacy `Notifier` interface directly, duplicating the equivalent `TrafficBreached`/`RecoveryStarted`/`RecoverySucceeded`/`RecoveryFailed` telemetry emission. Domain code now only emits telemetry - it never touches notifiers directly.
- New `Wiring::NotifierBridge` subscribes those four transition events on the telemetry bus and calls the `FailSafe`-wrapped notifiers from there. `LockChanged` is deliberately excluded (manual lock overrides were never legacy notifier events).

### DIFF
--- a/lib/stoplight/wiring/system.rb
+++ b/lib/stoplight/wiring/system.rb
@@ -65,6 +65,9 @@ module Stoplight
         @failover_system = failover_system
         @registry = registry
         @telemetry = Domain::Telemetry::Bus.new(error_notifier: config.error_notifier)
+
+        notifiers = wrapped_notifiers
+        NotifierBridge.new(notifiers:).subscribe(@telemetry) unless notifiers.empty?
       end
 
       def persistent?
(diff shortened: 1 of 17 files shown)


lib/stoplight/wiring/system.rb
```diff
@@ -65,6 +65,10 @@ def initialize(config:, failover_system:, registry:)
         @failover_system = failover_system
         @registry = registry
         @telemetry = Domain::Telemetry::Bus.new(error_notifier: config.error_notifier)
+
+        unless config.notifiers.empty?
```
Lokideos, line 69: `NIT`
I might be biased here, but can we use `#any?` here with `if`? 😅

```suggestion
        if config.notifiers.any?
```

Lokideos (approved):
Great work! 🚀 
Now our architecture docs are a little bit stale, but I think we can update them with the new version release 🙂

bolshakov:
yeah, honestly not sure what to do about docs - we have a fleet of third-party notifiers and probably they'd better stay as is

lib/stoplight/wiring/system.rb
bolshakov: I don't like added complexity of `#any?` - `#empty?` checks only the size of the list, while `#any?` inspect lists elements. It's basically a method to checking if any element matches a pattern with "truthy" being a default pattern

PR #920, opened 2026-09-02
Production Code: Use Redis Time in State & Metrics

bolshakov:
Updates production Lua scripts and Ruby code to use Redis-synchronized `now()` for all time-dependent operations:

1. Lua scripts call `now()` instead of duplicating time logic
   - State transitions (green→yellow→red) use Redis time
   - Metrics snapshots (`last_success_at`, `consecutive_errors`) use Redis time
2. Out-of-order protection removed ээ impossible with monotonic Redis time - time leaping backwards possible in theory but rather theoretical - in practice time adjustments is done by slowing down time.
3. Feature tests use real `sleep()` instead of time-travel to advance time with actual Redis calls
4. Cleanup removes unused `TimeTravel` from feature test helpers

 Part of: 3-phase refactoring (Phase 3/3)

### DIFF
--- a/features/stoplight/step_definitions/helpers_steps.rb
+++ b/features/stoplight/step_definitions/helpers_steps.rb
@@ -15,7 +15,7 @@ When(/^(\d+) (seconds|minutes|hours|days) have elapsed$/) do |seconds, unit|
   else
     raise ArgumentError, "Unknown time unit: #{unit}"
   end
-  Timecop.travel(Time.now + seconds)
+  sleep(seconds)
 end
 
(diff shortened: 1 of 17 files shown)


features/stoplight/step_definitions/helpers_steps.rb
```diff
@@ -15,7 +15,7 @@
   else
     raise ArgumentError, "Unknown time unit: #{unit}"
   end
-  Timecop.travel(Time.now + seconds)
+  sleep(seconds)
```
bolshakov, line 18: no time traveling in feature tests anymore

PR #921, opened 2026-09-02
Use Redis Time in State & Metrics to prevent time skew problem

bolshakov:
Updates production Lua scripts and Ruby code to use Redis-synchronized now() for all time-dependent operations:

1. Lua scripts call now() instead of duplicating time logic
  - Sta1te transitions (green→yellow→red) use Redis time
  - Metrics snapshots (last_success_at, consecutive_errors) use Redis time
  - Out-of-order protection removed ээ impossible with monotonic Redis time - time leaping backwards possible in theory but rather theoretical - in practice time adjustments is done by slowing down time.
2. Feature tests use real sleep() instead of time-travel to advance time with actual Redis calls
3. Cleanup removes unused TimeTravel from feature test helpers

### DIFF
--- a/lib/stoplight/infrastructure/redis/storage/state.rb
+++ b/lib/stoplight/infrastructure/redis/storage/state.rb
@@ -103,7 +103,7 @@ module Stoplight
           def transition_to_green
             became_green = scripting.call(
               "state/transition_to_green",
-              args: [clock.current_time.to_f],
+              args: [],
               keys: [state_key]
             )
             became_green == 1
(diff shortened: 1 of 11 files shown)


features/stoplight/global_configuration/window_size.feature
```diff
@@ -53,8 +53,8 @@ Feature: Stoplight configuration - Window Size
       | Threshold   | 3  |
     And the service starts failing with "timeout"
     And 1 request is made
-    And 1000 days have elapsed
+    And 1 hour have elapsed
     And 1 request is made
-    And 1000 days have elapsed
+    And 1 hour have elapsed
```
bolshakov, line 58: sleep 1 hour? WTF

PR #934, opened 2026-09-22
fix: Compute cool-off expiry from Redis

Lokideos:
One of the claims we make in #932 is "All time-dependent decisions are now made from Redis's own clock", which currently is wrong for calculating `recovery_scheduled_after_ts` and snapshot timestamps.

This PR aims to fix it by moving timestamp calculation to Redis.


lib/stoplight/infrastructure/redis/storage/state.rb
```diff
@@ -57,16 +57,19 @@ def set_state(state)
...
+                keys: [state_key]
+              )
 
             Domain::StateSnapshot.new(
               breached_at: breached_at_raw && clock.at(breached_at_raw.to_f),
               locked_state: locked_state&.to_sym || Stoplight::State::UNLOCKED,
               recovery_scheduled_after: recovery_scheduled_after_raw && clock.at(recovery_scheduled_after_raw.to_f),
               recovery_started_at: recovery_started_at_raw && clock.at(recovery_started_at_raw.to_f),
-              time: clock.current_time
+              time: clock.at(now_ms.to_i / 1000.0)
```
bolshakov: Is now ms a string? Maybe lets use fdiv here

